Issue #30, opened 2017-03-24
8184.10 DEV Add GET /db


a-murphy:
Shippable/pm#8184

Add `GET /db`
     - return the db information in the same format as vault right now
     - the values should be parsed from the env variables already set in the application

stephanielingwood:
verified with a new DO install

PR #55, opened 2017-03-29
Return an object from GET /vault

stephanielingwood:
https://github.com/Shippable/admiral/issues/30

Consistently return an object from GET /vault


stephanielingwood:
verified with a new install on DO

Issue #75, opened 2017-04-07
8312.2 Run system_configs.sql from /api/db/initialize


stephanielingwood:
Shippable/pm#8312

  - Write system_configs.sql. This should:
    - If there is no systemConfigs table, create the table with the default values
    - If a column doesn't exist, create that column
    - Columns: [systemConfigs.txt](https://github.com/Shippable/heap/files/904006/systemConfigs.txt)
  - Run this script in /api/db/initialize

PR #140, opened 2017-04-11
#75; adds systemConfigs to POST /api/db/initialize.

a-murphy:
#75 

Adds more defaults to the columns, removes the default systemConfigs.json, and moves the script to create systemConfigs to the API route.  Tested by calling the API route and then checking the systemConfigs with that API route.


common/scripts/configs/system_configs.sql
```diff
@@ -84,63 +74,51 @@ do $$
...
-      alter table "systemConfigs" add column "customHostDockerVersion" VARCHAR(24);
-    end if;
-
-    if not exists (select 1 from information_schema.columns where table_name = 'systemConfigs' and column_name = 'wwwPort') then
-      alter table "systemConfigs" add column "wwwPort" INTEGER;
-    end if;
-
-    if not exists (select 1 from information_schema.columns where table_name = 'systemConfigs' and column_name = 'redisUrl') then
-      alter table "systemConfigs" add column "redisUrl" varchar(255);
+      alter table "systemConfigs" add column "customHostDockerVersion" VARCHAR(24) default '1.9.1';
```
stephanielingwood: @ric03uec  - the current docker default in base's state.json.example is 1.9.1. What do we want the default to be in Admiral? Keep it 1.9.1, make it 1.13, or...???

common/scripts/create_sys_configs.sh
```diff
@@ -1,314 +1,8 @@
...
-  __process_msg "Updating : release"
-  sed -i "s#{{RELEASE}}#$RELEASE#g" $system_configs_sql
-
-  __process_msg "Successfully generated 'systemConfig' table data"
-}
-
 __copy_system_configs() {
   __process_msg "Copying systemConfigs.sql to db container"
-  local system_config_host_location="$CONFIG_DIR/system_configs.sql"
+  local system_config_host_location="/home/shippable/admiral/common/scripts/configs/system_configs.sql"
```
ric03uec: this should be set by setting env  `SCRIPTS_DIR` (see vault post.js).

Issue #84, opened 2017-04-07
8312.3 Vault: Add vault to the initialization flow


stephanielingwood:
Shippable/pm#8312

In `/api/workflow/initialize`...
  - call GET `/api/vault`. If vault is not initialized...
  - call POST /api/vault`

In the UI:
  - add vault to the table in the initialize panel

trriplejay:
should we be using GET `/api/vault`, or should we be using the systemConfig `secrets` object to determine if things are already initialized or not?

trriplejay:
finished.

PR #158, opened 2017-04-12
initialize vault in workflow

trriplejay:
#84

results in:
image

### DIFF
--- a/api/workflow/initialize.js
+++ b/api/workflow/initialize.js
@@ -23,6 +23,8 @@ function initialize(req, res) {
   async.series([
       _checkInputParams.bind(null, bag),
       _initializeDatabase.bind(null, bag),
+      _getSecrets.bind(null, bag),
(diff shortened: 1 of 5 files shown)


trriplejay:
Hmm. Left some extra stuff in the pr. Will fix in a bit.

ric03uec:
@trriplejay rebase required

ric03uec:
i think we should remove the cleanup file and add it as a command in installer. something like 
`./admiral.sh clean`
this should ask for prompt and remove containers, folders and any configs

api/workflow/initialize.js
```diff
@@ -53,3 +55,56 @@ function _initializeDatabase(bag, next) {
     }
   );
 }
+
+function _getSystemConfigs(bag, next) {
```
trriplejay: Will remove this.

Issue #120, opened 2017-04-07
8312.6 Add logs modal


stephanielingwood:
Shippable/pm#8312

Add log modals to the logs buttons in the UI.  Clicking the logs button for a step should open the log modal and display the logs for the correct step.

stephanielingwood:
complete. closing.

PR #179, opened 2017-04-12
add log modal support

trriplejay:
#120 

image


common/scripts/docker/installVault.sh
```diff
@@ -164,6 +164,7 @@ main() {
     __update_vault_config
     __update_vault_creds
     __run_vault
+    sleep 5
```
ric03uec: why is this required? can we not use the check_vault function ?

Issue #110, opened 2017-04-07
8312.5 Start fake API container


stephanielingwood:
Shippable/pm#8312

Add the following to POST `/api/db/initialize`:

Start an API container.  Give it the following environment variables (all from admiral.env):
  * DBNAME
  * DBUSERNAME
  * DBPASSWORD
  * DBHOST
  * DBPORT
  * DBDIALECT
  It will start, create the tables, and crash because it's missing some systemConfigs.

set the db to initialized after this is complete.

PR #181, opened 2017-04-13
#110; starts fake API container.

a-murphy:
#110 

Boxes for the access key and secret have been added.  They can't be hidden for now because we can't tell if the key and secret have been set from there, but they may be left empty.

screenshot from 2017-04-12 17 14 22

API does start, runs for about ten seconds, and is removed.


common/scripts/docker/startFakeAPI.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash -e
+
+source "$CONFIG_DIR/admiral.env"
+
+export COMPONENT="fakeapi"
```
ric03uec: this should be `db` so that all logs here are dumped on db console

common/scripts/docker/startFakeAPI.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash -e
+
+source "$CONFIG_DIR/admiral.env"
```
ric03uec: why do we need to source this? if any envs are required, they should be passed from the route (like CONFIG_DIR, SCRIPTS_DIR etc)

common/scripts/docker/startFakeAPI.sh
```diff
@@ -0,0 +1,118 @@
...
+  done
+
+  if [ $is_booted = false ]; then
+    __process_error "Failed to boot api container"
+    exit 1
+  fi
+}
+
+main() {
+  __process_marker "Booting api"
```
ric03uec: should be `Booting fake api to generate models`

common/scripts/configs/credentials.template
```diff
@@ -0,0 +1,3 @@
+[default]
+aws_access_key_id={{aws_access_key}}
```
ric03uec: lets rename the template variable to `{{AWS_ACCESS_KEY}}` to be consistent with variables in vault config template

ric03uec: in fact, if the admiral env file is calling it ACCESS_KEY, then it should be `{{ACCESS_KEY}}` here too

Issue #186, opened 2017-04-13
8312 return empty systemConfigs if not present


ric03uec:
- https://github.com/Shippable/pm/issues/8312
- after the first login, the api throws following error
![image](https://cloud.githubusercontent.com/assets/1110988/25018512/49560ba6-203c-11e7-81fa-85f7e25d6644.png)

- the get route for system configs should just return empty object/array if the table is not present instead of throwing an exception on the console

trriplejay:
@ric03uec 

hmm... the route today is returning an object for a successful get call.  Typically (with shippable api) if a single-object GET object does not exist yet, we return 404, not empty object.  Are you sure we want empty object, or should I just return a 404? UI can just ignore it.

The object you show above is actually an error from the postgres client itself because the systemConfigs table does not even exist yet, so basically I can make all errors result in a 404, but you'll probably still see this in the logs, unless we explicitly hide it.

stephanielingwood:
complete. closing.

PR #195, opened 2017-04-13
dont return error if sysconfigs arent added yet

trriplejay:
#186

if the systemConfigs query returns the error code that means the relation doesn't even exist yet, we will just return an empty object instead of throwing an error, since this means that just haven't clicked 'initialize' yet.

Any other error from the query will actually log an error as a DBOperationFailed.

if the query has no error, but no systemConfigs are found, then we return a 404 since this should never happen.


stephanielingwood:
could you also update the ui to stop ignoring errors from the get systemConfigs calls? we are currently ignoring errors, but won't need to anymore with this change.

Issue #199, opened 2017-04-13
8312 add GET /api/admiral route


ric03uec:
- https://github.com/Shippable/pm/issues/8312
- add route `GET /api/admiral`
- the handler should return the values in `admiral.env` in json format.

stephanielingwood:
complete. closing.

PR #209, opened 2017-04-14
#199; adds GET /api/admiral.

a-murphy:
#199 

Returns an object with the keys and values from admiral.env.

### DIFF
--- a/api/admiral/get.js
+++ b/api/admiral/get.js
@@ -1,0 +1,90 @@
+'use strict';
+
+var self = get;
+module.exports = self;
+
+var async = require('async');
+var _ = require('underscore');
+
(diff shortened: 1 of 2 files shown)


api/admiral/get.js
```diff
@@ -0,0 +1,66 @@
...
+    function (envLine) {
+      if (!_.isEmpty(envLine) && envLine.indexOf('=') >= 0) {
+        var name = envLine.split('=')[0];
+        var value = envLine.split('=')[1];
+        bag.resBody[name] = JSON.parse(value);
+      }
+    }
+  );
+
+  filereader.on('close',
```
ric03uec: needs error handling here

Issue #218, opened 2017-04-16
8344.1 Add POST /api/providers


a-murphy:
Shippable/pm#8344

Add a `POST /api/providers` route, copying the logic in API's `POST /providers`.

stephanielingwood:
complete. closing.

PR #242, opened 2017-04-16
Add POST providers

stephanielingwood:
https://github.com/Shippable/admiral/issues/218

- Adds POST providers route using the logic from api

### DIFF
--- a/api/providers/post.js
+++ b/api/providers/post.js
@@ -1,0 +1,158 @@
+'use strict';
+
+var self = post;
+module.exports = self;
+
+var async = require('async');
+var mongoose = require('mongoose');
+var _ = require('underscore');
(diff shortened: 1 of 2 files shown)


api/providers/post.js
```diff
@@ -0,0 +1,158 @@
...
+      new ActErr(who, ActErr.DataNotFound,
+        'Missing body data or wrong type :url')
+    );
+
+  if (!bag.reqBody.name)
+    return next(
+      new ActErr(who, ActErr.DataNotFound, 'Missing body data :name')
+    );
+
+  // This is a special case for azure as its dns does
```
a-murphy: Maybe make this comment a little broader?  It can't be the only case.

Issue #223, opened 2017-04-16
8344.1 Add GET /api/systemIntegrations


a-murphy:
Shippable/pm#8344

Add a `GET /api/systemIntegrations` route.  This will accept a `name` query parameter.  If there is a `name` in the query string, only that integration is returned.  Otherwise all system integrations are returned.

PR #244, opened 2017-04-16
#223; adds GET /api/systemIntegrations.

a-murphy:
#223 

Adds `GET /api/systemIntegrations`, which allows queries by `name` and `masterName`.


stephanielingwood:
lgtm

Issue #219, opened 2017-04-16
8344.1 Add GET /api/providers


a-murphy:
Shippable/pm#8344

Add a `GET /api/providers` route, copying the logic in API's `GET /providers`.

stephanielingwood:
complete. closing.

PR #245, opened 2017-04-16
Adds GET providers route

stephanielingwood:
https://github.com/Shippable/admiral/issues/219

- Adds GET providers, with the same logic as in API
- Can query by providerIds, providerUrls, or providerUrlSlugs

### DIFF
--- a/api/providers/get.js
+++ b/api/providers/get.js
@@ -1,0 +1,107 @@
+'use strict';
+
+var self = get;
+module.exports = self;
+
+var async = require('async');
+var _ = require('underscore');
+
+function get(req, res) {
+  var bag = {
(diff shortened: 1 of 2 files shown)


api/providers/get.js
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  bag.who = util.format('providers|%s', self.name);
+  logger.info(bag.who, 'Starting');
+
+  async.series([
+    _checkInputParams.bind(null, bag),
+    _constructQuery.bind(null, bag),
+    _getS.bind(null, bag)
+  ],
```
a-murphy: Indentation.

Issue #233, opened 2017-04-16
8344.3 Add POST /api/services


a-murphy:
Shippable/pm#8344

Add `POST /api/services`:
  * accepts the following request body: `{name: <service name, replicas:<replica count>}`
  * generates the service specific envs in handler
  * stops any existing service with same name
  * starts a new service using docker command

PR #262, opened 2017-04-18
add post services route

ric03uec:
- https://github.com/Shippable/admiral/issues/233

### DIFF
--- a/api/services/post.js
+++ b/api/services/post.js
@@ -104,7 +104,8 @@ function _generateServiceConfig(bag, next) {
   logger.verbose(who, 'Inside');
 
   var configGenerator = null;
-  if (bag.name === 'api') configGenerator = apiConfig;
+  if (bag.name === 'api')
+    configGenerator = apiConfig;
 
   var params = {
     config: bag.serviceConfig,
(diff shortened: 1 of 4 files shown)


api/services/post.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+  );
+}
+
+function _generateServiceConfig(bag, next) {
+  var who = bag.who + '|' + _generateServiceConfig.name;
+  logger.verbose(who, 'Inside');
+
+  var configGenerator = null;
+  if (bag.name === 'api') configGenerator = apiConfig;
```
a-murphy: Wouldn't an object to map names to generators here be easier to work with?

stephanielingwood: this should be two lines

common/scripts/boot_service.sh
```diff
@@ -0,0 +1,40 @@
...
+export LOGS_FILE="$RUNTIME_DIR/logs/$SERVICE_NAME.log"
+
+## Write logs of this script to component specific file
+exec &> >(tee -a "$LOGS_FILE")
+
+__validate_service_configs() {
+  __process_msg "Service $SERVICE_NAME configuration"
+  __process_msg "SERVICE: $SERVICE_NAME"
+  __process_msg "LOGS_FILE:$LOGS_FILE"
+
```
stephanielingwood: remove newline

api/services/post.js
ric03uec: for all the microservices, the name will map to same handler since they all have same configs(env, mounts etc). so this should work for now. if the logic gets complicated, we can add a map in next heap issue

Issue #264, opened 2017-04-18
8344.3 Add GET /api/services


ric03uec:
- Shippable/pm#8344

- Add `GET /api/services?name=<foo>`:
    - accepts query param to return the service with name
    - if no query param specified, return all services
    - if no services pesent in DB, then parse `services/services.json` file to create services array in system settings for core services
    - the format for `services` object in `systemSettings` is
```
{
    "api": { "serviceName": "api", "mounts": "", "envs": "", "opts": "", "replicas": "global"},
    "www": { "serviceName": "www", "mounts": "", "envs": "", "opts": "", "replicas": "global"},
    "mktg": { "serviceName": "mktg", "mounts": "", "envs": "", "opts": "", "replicas": "global"},
    "sync": { "serviceName": "sync", "mounts": "", "envs": "", "opts": "", "replicas": 1}
}
```

stephanielingwood:
complete. closing.

PR #271, opened 2017-04-19
Adds GET /api/services; moves services.json to /common/scripts/configs

stephanielingwood:
https://github.com/Shippable/admiral/issues/264

- Moves services.json to /common/scripts/configs; removes envs and volumes from the serviceConfigs
- Removes $SCRIPTS_DIR
- Adds GET /api/services. This will:
  - If nothing is present in systemSettings.services, reads the services.json and adds any core services (those with `isCore: true`) to the systemSettings.services object
  - If no query string is specified, returns an array of services
  - For a query string `?name=api`, will return an array with the requested service
  - If the requested service is not found in the services object, returns an empty array


api/services/get.js
```diff
@@ -0,0 +1,152 @@
...
+  );
+
+  configHandler.put(bag.component, bag.defaultService,
+    function (err) {
+      if (err)
+        return next(
+          new ActErr(who, ActErr.OperationFailed, err)
+        );
+
+      bag.resBody = bag.defaultService;
```
a-murphy: What if this route was called with a queryString and the default had not been set?

api/services/get.js
```diff
@@ -0,0 +1,152 @@
...
+        );
+
+      if (_.isEmpty(services)) {
+        logger.debug('No configuration in database for ' + bag.component);
+        bag.initializeDefault = true;
+        return next();
+      }
+
+      if (bag.serviceQuery) {
+        if (!services[bag.serviceQuery])
```
ric03uec: `_.has(services, bag.serviceQuery)` will be more readable here

api/services/get.js
```diff
@@ -0,0 +1,152 @@
...
+          return next(
+            new ActErr(who, ActErr.DataNotFound,
+              util.format('No service \'%s\' found in systemSettings.services',
+                bag.serviceQuery), err)
+          );
+        bag.resBody = services[bag.serviceQuery];
+
+      } else {
+        bag.resBody = services;
```
ric03uec: bag.resBody should be an array and it should return either a single service or all services based on query param

api/services/get.js
```diff
@@ -0,0 +1,152 @@
...
+  };
+
+  bag.who = util.format('services|%s', self.name);
+  logger.info(bag.who, 'Starting');
+
+  async.series([
+    _checkInputParams.bind(null, bag),
+    _get.bind(null, bag),
+    _getServicesJson.bind(null, bag),
+    _setDefault.bind(null, bag)
```
ric03uec: the `_get` should be called at the end. it should return values from database after they've been set. so the workflow should be something like 
- check if db has any values
- if not, set default values
- return whatever is in the database
subsequent calls to this route should skip steps 1 and 2 and just get value from db

Issue #274, opened 2017-04-19
8344.3 Add POST /api/services handler for www


stephanielingwood:
https://github.com/Shippable/pm/issues/8344

stephanielingwood:
complete. closing.

PR #278, opened 2017-04-20
Add www handler to POST /api/services

stephanielingwood:
https://github.com/Shippable/admiral/issues/274

- Adds a handler to bring up www
- Moves getting vault creds into the apiConfig file

### DIFF
--- a/api/services/post.js
+++ b/api/services/post.js
@@ -9,22 +9,23 @@ var _ = require('underscore');
 var spawn = require('child_process').spawn;
 var fs = require('fs');
 
+var APIAdapter = require('../../common/APIAdapter.js');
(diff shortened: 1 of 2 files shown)


api/services/apiConfig.js
```diff
@@ -43,6 +50,57 @@ function _checkInputParams(bag, next) {
...
+
+  envHandler.get(bag.vaultTokenEnv,
+    function (err, value) {
+      if (err)
+        return next(
+          new ActErr(who, ActErr.OperationFailed,
+            'Cannot get env: ' + bag.vaultTokenEnv)
+        );
+
+      if (_.isEmpty(value)) {
```
a-murphy: You don't need these brackets.

Issue #307, opened 2017-04-24
8385.1 Add Bitbucket to the UI


a-murphy:
Shippable/pm#8385

Add a section for Bitbucket to the UI with the current GitHub client section

stephanielingwood:
complete. closing.

PR #355, opened 2017-04-24
Add bitbucket as an auth provider

stephanielingwood:
https://github.com/Shippable/admiral/issues/307

- Adds bitbucket as an auth provider

bitbucket

### DIFF
--- a/static/scripts/dashboard/dashboardCtrl.js
+++ b/static/scripts/dashboard/dashboardCtrl.js
@@ -34,71 +34,98 @@
         accessKey: '',
         secretKey: ''
       },
+      // map by systemInt name, then masterName
       installForm: {
         api: {
-          isEnabled: true,
-          masterName: 'url',
-          data: {
-            url: ''
+          url: {
+            isEnabled: true,
+            masterName: 'url',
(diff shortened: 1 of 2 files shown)


static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -787,6 +834,16 @@
           if (err)
             return next(err);
 
+          if (bag.name === 'auth') {
+            // remove 'Keys' off the masterName to
+            // get the key name for the installForm
+            var key = bag.systemIntegration.masterName.substr(0,
```
ric03uec: - we cannot do this. systemIntegrations are not being translated here correctly because of which such logic is being added. the `name` field in `systemIntegrations` is **not** the primary key and the logic in this file assumed it was. 
- the `$scope.vm.installForm` variable should create an array (or a map) for each integration name. so it should be something like
```
$scope.vm.installForm = {
"api": [{masterName: url, data: {}}]
"auth": [{masterName: githubKeys, data{}}, {masterName: bitbucketKeys, data: {}}],
}
```

Issue #320, opened 2017-04-24
8385.7 Add GET /api/masterIntegrations


a-murphy:
Shippable/pm#8385

Add `GET /api/masterIntegrations` to Admiral.

PR #357, opened 2017-04-24
adding GET /api/masterIntegrations route

ric03uec:
- https://github.com/Shippable/admiral/issues/320

### DIFF
--- a/api/masterIntegrations/get.js
+++ b/api/masterIntegrations/get.js
@@ -10,7 +10,7 @@ function get(req, res) {
   var bag = {
     inputParams: req.params,
     reqBody: req.body,
-    resBody: {}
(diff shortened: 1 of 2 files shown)


api/masterIntegrations/get.js
```diff
@@ -0,0 +1,66 @@
...
+
+  var query = 'SELECT * FROM "masterIntegrations";';
+
+  global.config.client.query(query,
+    function (err, masterIntegrations) {
+      if (err)
+        return next(
+          new ActErr(who, ActErr.DBOperationFailed, err)
+        );
+      if (_.isEmpty(masterIntegrations.rows))
```
a-murphy: Why not return an empty array?

Issue #407, opened 2017-04-26
8385 Disabling dyamic nodes is not removing ec2 and marshaller


ric03uec:
- https://github.com/Shippable/pm/issues/8385
- when dynamic nodes integration is disabled from admin panel, ec2 and marshaller services should be removed 
- marshaller is also not started when the `services` column is reset in the db and user clicks `Install` from admin panel

stephanielingwood:
two changes:
- In DELETE /api/systemIntegrations, disable the masterIntegration only if there are no other systemIntegrations using that master integration
- Remove logup from core services; it should be started when the amazonKeys masterInt is enabled
- In the addons panel, call a PUT on every masterintegration when the addons button is clicked. (Right now the masterIntegration PUT is only being called when it has changed.)

stephanielingwood:
complete. closing.

PR #413, opened 2017-04-27
Disable masterIntegration when the last systemInt using it is deleted

stephanielingwood:
https://github.com/Shippable/admiral/issues/407

- When the last systemInt using a masterInt is deleted, disable that masterInt
- Call a put on the masterInt to enable the masterInt when PUT systemInt is called (this handle the case where the services have been deleted from the db, and the user clicks install; all previously existing services managed by the install button will be brought back up)
- Removes logup from core services; it is now booted when amazonKeys is enabled
- Calls PUT /masterIntegration for all masterInts in the addons panel each time the "Install Addons" button is clicked

### DIFF
--- a/api/systemIntegrations/deleteById.js
+++ b/api/systemIntegrations/deleteById.js
@@ -5,6 +5,7 @@ module.exports = self;
 
 var async = require('async');
 var _ = require('underscore');
+var APIAdapter = require('../../common/APIAdapter.js');
 var envHandler = require('../../common/envHandler.js');
 var VaultAdapter = require('../../common/VaultAdapter.js');
 
@@ -12,6 +13,7 @@ function deleteById(req, res) {
   var bag = {
     inputParams: req.params,
     resBody: {},
+    apiAdapter: new APIAdapter(req.headers.authorization.split(' ')[1]),
     vaultUrlEnv: 'VAULT_URL',
     vaultTokenEnv: 'VAULT_TOKEN'
   };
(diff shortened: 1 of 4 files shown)


api/systemIntegrations/deleteById.js
```diff
@@ -50,6 +55,28 @@ function _checkInputParams(bag, next) {
...
+  logger.verbose(who, 'Inside');
+
+  var query = util.format('SELECT * FROM "systemIntegrations" WHERE id=\'%s\'',
+    bag.systemIntegrationId);
+
+  global.config.client.query(query,
+    function (err, systemIntegrations) {
+      if (err)
+        return next(
+          new ActErr(who, ActErr.DBOperationFailed, err)
```
a-murphy: Could you put a message in the ActErr?

api/systemIntegrations/deleteById.js
```diff
@@ -147,3 +174,53 @@ function _deleteByIdInVault(bag, next) {
...
+  var who = bag.who + '|' + _getSystemIntegrations.name;
+  logger.verbose(who, 'Inside');
+
+  var query = 'SELECT * FROM "systemIntegrations"';
+
+  global.config.client.query(query,
+    function (err, systemIntegrations) {
+      if (err)
+        return next(
+          new ActErr(who, ActErr.DBOperationFailed, err)
```
a-murphy: And here?

static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -965,7 +964,6 @@
           updateSMTPSystemIntegration,
           enableEmailMasterIntegration,
```
a-murphy: Could you take the line to skip this update if it's already enabled out, too?

Issue #421, opened 2017-04-30
8405.3 add db cleanup route


ric03uec:
- https://github.com/Shippable/pm/issues/8405
- add route `POST /api/db/cleanup`. this should run post installation scripts

PR #432, opened 2017-04-30
adding POST db/cleanup route

ric03uec:
- https://github.com/Shippable/admiral/issues/421

### DIFF
--- a/common/scripts/post_install.sh
+++ b/common/scripts/post_install.sh
@@ -1,0 +1,47 @@
+#!/bin/bash -e
+
+export COMPONENT="db"
+export DB_DATA_DIR="$RUNTIME_DIR/$COMPONENT/data"
+export DB_CONFIG_DIR="$CONFIG_DIR/$COMPONENT"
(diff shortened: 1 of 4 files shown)


common/scripts/post_install.sh
```diff
@@ -0,0 +1,47 @@
...
+  __process_msg "LOGS_FILE:$LOGS_FILE"
+}
+
+__copy_migrations() {
+  __process_msg "Copying post install sql to db container"
+  local migrations_host_location="$MIGRATIONS_DIR/post_install/post_install.sql"
+  local migrations_container_location="$CONFIG_DIR/db/post_install.sql"
+  sudo cp -vr $migrations_host_location $migrations_container_location
+
+  __process_msg "Successfully copied migrations.sql to db container"
```
a-murphy: Could you update this log to refer to the correct file?

common/scripts/post_install.sh
```diff
@@ -0,0 +1,47 @@
...
+}
+
+__migrate() {
+  __process_msg "Running migrations"
```
a-murphy: And maybe update this, too?

Issue #461, opened 2017-05-02
8405 Add kernel requirements to README


stephanielingwood:
Shippable/pm#8405

- In the instructions, add  the command to update the kernel to 3.16. Only AWS users need to do this.
- Also add a "known issues" section and say that admiral doesn't run on kernels < 3.16.

PR #463, opened 2017-05-02
updating docs

ric03uec:
- https://github.com/Shippable/admiral/issues/461

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,14 +24,35 @@
     - 8 Gb memory
     - 100 Gb disk
     - Ubuntu 14.04 LTS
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,6 +46,15 @@ $ sudo apt-get update
 $ sudo apt-get install git-core
 ```
 
+#### Upgrade kernel
+- If installer is running on AWS ubuntu 14.04 ami, then the kernel needs to be
+  updated. Run following commands to update kernel to 3.19
```
a-murphy: Don't you also need to restart the instance after updating the kernel?

Issue #472, opened 2017-05-04
8413.1 Add database status route


stephanielingwood:
https://github.com/Shippable/pm/issues/8413

Add a database status route.

PR #532, opened 2017-05-05
adding db status route

ric03uec:
- https://github.com/Shippable/admiral/issues/472

### DIFF
--- a/api/db/getStatus.js
+++ b/api/db/getStatus.js
@@ -9,7 +9,10 @@ var pg = require('pg');
 function getStatus(req, res) {
   var bag = {
     reqQuery: req.query,
-    resBody: {}
+    resBody: {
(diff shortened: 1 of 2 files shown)


api/db/getStatus.js
```diff
@@ -0,0 +1,68 @@
...
+    global.config.dbUsername,
+    global.config.dbPassword,
+    global.config.dbHost,
+    global.config.dbPort,
+    global.config.dbName);
+
+  var testClient = new pg.Client(connectionString);
+  testClient.connect(
+    function (err) {
+      logger.error(util.inspect(arguments))
```
a-murphy: Will this be a helpful message?  It could use some sort of label.  And we probably don't want to log it if it's successful.

Issue #485, opened 2017-05-04
8413.3 Add a status route for Vault


stephanielingwood:
Shippable/pm#8413

Add a status route that checks whether Vault is accepting connections

PR #546, opened 2017-05-06
adding status route for vault

ric03uec:
- https://github.com/Shippable/admiral/issues/485

### DIFF
--- a/api/secrets/getStatus.js
+++ b/api/secrets/getStatus.js
@@ -89,7 +89,7 @@ function _getStatus(bag, next) {
     function (err, response) {
       if (err && err.code === 'ECONNREFUSED') {
         bag.resBody.isReachable = false;
-        bag.resBody = _.extend(bag.resBody, err);
+        bag.resBody.error = util.inspect(err);
       } else if (err) {
         return next(
           new ActErr(who, ActErr.OperationFailed,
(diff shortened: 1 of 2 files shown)


api/secrets/getStatus.js
```diff
@@ -0,0 +1,106 @@
...
+  var who = bag.who + '|' + _getStatus.name;
+  logger.verbose(who, 'Inside');
+
+  bag.vaultAdapter.getHealth(
+    function (err, response) {
+      if (err && err.code === 'ECONNREFUSED') {
+        bag.resBody.isReachable = false;
+        bag.resBody.error = util.inspect(err);
+      } else if (err) {
+        return next(
```
a-murphy: What makes this on an ActErr and the other one a `resBody.error`?

api/secrets/getStatus.js
```diff
@@ -0,0 +1,105 @@
...
+  var who = bag.who + '|' + _getStatus.name;
+  logger.verbose(who, 'Inside');
+
+  bag.vaultAdapter.getHealth(
+    function (err, response) {
+      if (err && err.code === 'ECONNREFUSED') {
+        bag.resBody.isReachable = false;
+        bag.resBody.error = util.inspect(err);
+      } else if (err) {
+        bag.resBody.isReachable = true;
```
a-murphy: Should this be `true`?

ric03uec: yes. the case for vault not being reachable is handled in the first error. at this stage, vault should be reachable

Issue #509, opened 2017-05-04
8413.8 Prompt if user would like to install with existing DB


stephanielingwood:
Shippable/pm#8413

Before asking for the database IP address, ask if the user would like to install a new database or use an existing one. If the user selects to use an existing database, don't install a new one.

PR #558, opened 2017-05-08
#509; asks if there is an existing database.

a-murphy:
#509 

Asks if the database will be new or existing, and skips installation for existing.

### DIFF
--- a/common/scripts/lib/_parseArgs.sh
+++ b/common/scripts/lib/_parseArgs.sh
@@ -85,6 +85,7 @@ __validate_runtime() {
   ################## check db url ##################################
   if [ "$DB_IP" == "" ]; then
     __process_msg "DB_IP is not set"
+    __set_install_db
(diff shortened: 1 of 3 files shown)


common/scripts/lib/_parseArgs.sh
```diff
@@ -85,6 +85,7 @@ __validate_runtime() {
   ################## check db url ##################################
   if [ "$DB_IP" == "" ]; then
     __process_msg "DB_IP is not set"
+    __set_install_db
```
ric03uec: this means that installer is asking more questions that it should for a completely onebox installation. the flow should be something like this
- ask the IP address of db (type D to use defautl)
    - if default is chosen, move to next step
- if some IP is provided, __then__ ask if a new installation is required or an existing one should be used
   - for new installation, run scripts
   - for existing ones, connect

Issue #501, opened 2017-05-04
8413.7 Add GET /api/workers route


stephanielingwood:
Shippable/pm#8413

add GET /api/workers route

    this should return the value for swarm workers and set default value if
    the workers column is empty in systemSettings

PR #564, opened 2017-05-08
adding workers GEt route

ric03uec:
- https://github.com/Shippable/admiral/issues/501

### DIFF
--- a/common/configHandler.js
+++ b/common/configHandler.js
@@ -71,7 +71,7 @@ function _getConfigs(bag, next) {
             'Failed to get config for component: ' + bag.component, err)
         );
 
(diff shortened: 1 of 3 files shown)


common/configHandler.js
```diff
@@ -71,7 +71,9 @@ function _getConfigs(bag, next) {
             'Failed to get config for component: ' + bag.component, err)
         );
 
-      bag.update = _.extend(config, bag.update);
+      if (!_.isArray(bag.update)) {
+        bag.update = _.extend(config, bag.update);
+      }
```
a-murphy: You don't need the brackets.

Issue #511, opened 2017-05-04
8413.8 Test that creds and connection info is valid


stephanielingwood:
Shippable/pm#8413

Create and drop a test table to check that the credentials are valid.

PR #565, opened 2017-05-08
#511; checks existing database credentials.

a-murphy:
#511 

Creates and drops "testTable" for existing databases.  And removes the option to set the database username.


common/scripts/installDb.sh
```diff
@@ -59,6 +59,26 @@ __install_db() {
...
+      -U $DB_USER \
+      -d $DB_NAME \
+      -h $DB_IP \
+      -p $DB_PORT \
+      -v ON_ERROR_STOP=1 \
+      -c 'DROP TABLE \"testTable\";'"
+     __process_msg "Creating a test table"
+     eval "$create_table_cmd"
+     __process_msg "Dropping test table"
+     eval "$drop_table_cmd"
```
ric03uec: lets add some logs here to tell users exactly what the installer did here. something like
- `creating a test table to verify connetion`
- `dropping a test table`
- `'successfully tested connection to <db ip> <db port>`

Issue #572, opened 2017-05-09
8413.7 add POST /api/worker/initialize route


ric03uec:
- https://github.com/Shippable/pm/issues/8413

PR #573, opened 2017-05-09
adding scripts to initialize swarm worker

ric03uec:
- https://github.com/Shippable/admiral/issues/572

### DIFF
--- a/api/workers/initialize.js
+++ b/api/workers/initialize.js
@@ -1,0 +1,410 @@
+'use strict';
+
+var self = initialize;
+module.exports = self;
+
+var async = require('async');
+var _ = require('underscore');
+var path = require('path');
+var fs = require('fs');
+var spawn = require('child_process').spawn;
+
+var configHandler = require('../../common/configHandler.js');
+var envHandler = require('../../common/envHandler.js');
+var APIAdapter = require('../../common/APIAdapter.js');
+
+function initialize(req, res) {
+  var bag = {
+    reqBody: req.body,
+    res: res,
+    resBody: {},
+    component: 'workers',
+    apiAdapter: new APIAdapter(req.headers.authorization.split(' ')[1]),
+    params: {},
+    skipStatusChange: false,
+    isResponseSent: false,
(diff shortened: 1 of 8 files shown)


api/workers/initialize.js
```diff
@@ -0,0 +1,410 @@
...
+
+  if (!_.has(bag.config, 'address') || _.isEmpty(bag.config.address))
+    missingConfigFields.push('address');
+  if (!_.has(bag.config, 'port') || !bag.config.port)
+    missingConfigFields.push('port');
+
+  if (missingConfigFields.length)
+    return next(
+      new ActErr(who, ActErr.DataNotFound,
+        'Missing config data: ' + missingConfigFields.join())
```
a-murphy: How about "Missing worker config data"?  It might not be clear that it's the worker config and not the master.

api/workers/initialize.js
```diff
@@ -0,0 +1,410 @@
...
+
+  return next();
+}
+
+function _setProcessingFlag(bag, next) {
+  var who = bag.who + '|' + _setProcessingFlag.name;
+  logger.verbose(who, 'Inside');
+
+  var update = {
+    isProcessing: true,
```
a-murphy: Do we only have one `isProcessing` for all workers?  (And one `isInitialized`?)

common/scripts/installWorker.sh
```diff
@@ -0,0 +1,103 @@
...
+export WORKERS_CONFIG_DIR="$CONFIG_DIR/$COMPONENT"
+export SCRIPTS_DIR="$SCRIPTS_DIR"
+export LOGS_FILE="$RUNTIME_DIR/logs/$COMPONENT.log"
+export SCRIPTS_DIR_REMOTE="/tmp/shippable"
+
+## Write logs of this script to component specific file
+exec &> >(tee -a "$LOGS_FILE")
+
+__validate_worker_envs() {
+  __process_msg "Initializing redis environment variables"
```
a-murphy: Is this Redis?

Issue #516, opened 2017-05-04
8413.9 Add handler to connect to existing vault


stephanielingwood:
Shippable/pm#8413

Add a handler to POST /api/secrets/initialize to connect to an existing vault instance
This should connect to the instance, and test that the credentials are valid by creating a key-value pair in vault, and then deleting it

PR #574, opened 2017-05-09
#516; adds credential validation for existing Vault.

a-murphy:
#516 

Checks the credentials by creating and deleting a key-value pair for existing Vault installations.  Tested with a new installation and reusing that Vault container in another Admiral installation.

### DIFF
--- a/api/secrets/post.js
+++ b/api/secrets/post.js
@@ -77,6 +77,9 @@ function _post(bag, next) {
   if (_.has(bag.reqBody, 'rootToken'))
     bag.config.rootToken = bag.reqBody.rootToken;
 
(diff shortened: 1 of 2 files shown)


api/secrets/post.js
```diff
@@ -77,6 +77,9 @@ function _post(bag, next) {
   if (_.has(bag.reqBody, 'rootToken'))
     bag.config.rootToken = bag.reqBody.rootToken;
 
+  if (_.has(bag.reqBody, 'isShippableManaged'))
+    bag.config.isShippableManaged = bag.reqBody.isShippableManaged;
```
ric03uec: what if the `isShippableManged` flag is not passed, are we setting this to a default value somewhere?

Issue #593, opened 2017-05-09
8413.7 boot service should create docker swarm service


ric03uec:
- https://github.com/Shippable/pm/issues/8413
- if an external swarm worker is initialized, then booting a service should bring up the service on the swarm cluster instead of admiral node

PR #595, opened 2017-05-09
booting services differently for swarm cluster and onebox

ric03uec:
- https://github.com/Shippable/admiral/issues/593

### DIFF
--- a/api/services/post.js
+++ b/api/services/post.js
@@ -38,6 +38,8 @@ function post(req, res) {
       _checkInputParams.bind(null, bag),
       _getReleaseVersion.bind(null, bag),
       _getServiceConfig.bind(null, bag),
+      _getMaster.bind(null, bag),
+      _getWorkers.bind(null, bag),
       _getAccessKey.bind(null, bag),
       _getSecretKey.bind(null, bag),
       _getRegistry.bind(null, bag),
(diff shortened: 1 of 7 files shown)


api/services/post.js
```diff
@@ -254,7 +315,7 @@ function _generateInitializeEnvs(bag, next) {
     'SERVICE_MOUNTS': bag.serviceConfig.mounts,
```
a-murphy: I can't comment on line 314, but does anything still have `opts`?

common/scripts/boot_service.sh
```diff
@@ -42,7 +26,7 @@ __cleanup_service() {
   sudo docker service rm $SERVICE_NAME || true
 }
 
-__run_service() {
+__run_service1() {
```
a-murphy: Is there a better name for this?

api/services/post.js
```diff
@@ -134,6 +136,65 @@ function _getServiceConfig(bag, next) {
...
+      if (_.isEmpty(workers))
+        return next(
+          new ActErr(who, ActErr.DataNotFound,
+            'No configuration in database for ' + bag.component)
+        );
+
+      bag.isSwarmClusterInitialized = false;
+
+      // if even one worker has external IP then swarm cluster is initialized
+      _.each(bag.workers,
```
stephanielingwood: can use a _.some here, if you only need to find one worker whose IP doesn't match the master

Issue #506, opened 2017-05-04
8413.7 Add GET /api/workers/logs


stephanielingwood:
Shippable/pm#8413

Add a route to get the logs from bringing up the swarm workers

PR #606, opened 2017-05-10
adding logs route for workers

ric03uec:
- https://github.com/Shippable/admiral/issues/506

### DIFF
--- a/api/workers/getLogs.js
+++ b/api/workers/getLogs.js
@@ -1,0 +1,96 @@
+'use strict';
+
+var self = getLogs;
+module.exports = self;
+
+var async = require('async');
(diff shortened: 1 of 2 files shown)


api/workers/getLogs.js
```diff
@@ -0,0 +1,96 @@
...
+    resBody: [],
+    component: 'workers',
+    skipRead: false
+  };
+
+  bag.who = util.format('workers|%s', self.name);
+  logger.info(bag.who, 'Starting');
+
+  async.series([
+    _checkInputParams.bind(null, bag),
```
a-murphy: This is usually indented two more spaces.

Issue #602, opened 2017-05-10
8413.13 Migration - populate core services data


stephanielingwood:
Shippable/pm#8413

- for each service in system settings (db, msg, rabbitmq etc) find the value from state file and update in sql template

- for services that need systemIntegrations, create a systemIntegration. Add this to the sql template.
- run that templated sql file.

PR #610, opened 2017-05-12
#602; adds configs and systemIntegrations to migration.

a-murphy:
#602 

Adds the configs and systemIntegrations to the migration.

### DIFF
--- a/baseMigrations/migrate.js
+++ b/baseMigrations/migrate.js
@@ -11,7 +11,13 @@ var fs = require('fs-extra');
 var pg = require('pg');
 var spawn = require('child_process').spawn;
 var _ = require('underscore');
-var util = require('util');
+var url = require('url');
+
+global.util = require('util');
+
+var VaultAdapter = require('../common/VaultAdapter.js');
+var createOrUpdateSystemIntegration =
+  require('./createOrUpdateSystemIntegration.js');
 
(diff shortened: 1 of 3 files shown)


baseMigrations/migrate.js
```diff
@@ -321,6 +346,626 @@ function __getValueForSystemSettings(bag, name, defaultValue) {
...
+      return machine.group === 'core' &&
+        (machine.name === 'swarm' || machine.name === 'localhost');
+    }
+  );
+
+  var master = {
+    address: swarmMaster.ip,
+    workerJoinToken: bag.stateJson.systemSettings.swarmWorkerToken || '',
+    port: 2377,
+    isInstalled: false,
```
stephanielingwood: these two should be true

baseMigrations/migrate.js
```diff
@@ -321,6 +346,626 @@ function __getValueForSystemSettings(bag, name, defaultValue) {
...
+  _.each(bag.stateJson.machines,
+    function (machine) {
+      if (machine.group !== 'services')
+        return;
+
+      workers.push({
+        address: machine.ip,
+        port: 2377,
+        name: machine.name,
+        isInstalled: false,
```
stephanielingwood: these two should be true

Issue #612, opened 2017-05-12
8413.7 Enforce uniqueness of worker names


a-murphy:
Shippable/pm#8413

Don't allow two Swarm workers to have the same name.

stephanielingwood:
complete. closing.

PR #615, opened 2017-05-12
Enforce uniqueness of swarm worker names

stephanielingwood:
https://github.com/Shippable/admiral/issues/612

- Validates that the name being added doesn't already exist in the workers array
- Also shows swarm worker init logs

### DIFF
--- a/api/workers/post.js
+++ b/api/workers/post.js
@@ -93,6 +93,13 @@ function _updateWorker(bag, next) {
     port: 2377
   };
 
+  var nameInUse = _.findWhere(bag.config, {name: bag.workerName});
+  if (nameInUse && (nameInUse.address !== bag.workerAddress))
+    return next(
+      new ActErr(who, ActErr.OperationFailed,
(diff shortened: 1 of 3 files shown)


static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -2423,6 +2437,24 @@
       );
     }
 
+    function showWorkersLogModal(service) {
+      $scope.vm.selectedService = {};
+      $scope.vm.selectedService.serviceName = service;
+      $scope.vm.selectedService.displayName = 'Swarm Worker ';
```
a-murphy: Should there be a space at the end of the display name?

api/workers/post.js
```diff
@@ -96,7 +96,7 @@ function _updateWorker(bag, next) {
   var workerExists = false;
   _.each(bag.config,
     function (worker) {
-      if (worker.address === bag.workerAddress) {
+      if (worker.name === bag.workerName) {
```
a-murphy: Address needs to be unique, too.

Issue #589, opened 2017-05-09
8413.15 add section to show super users in UI


ric03uec:
- https://github.com/Shippable/pm/issues/8413
- add a section in addons titled `System super users`
- this should show an empty text box followed by a list of accounts that are returned by `GET /api/superusers` route
- when user clicks `Install`, the value in text box is POST-ed in `/api/superusers` route if it is non-empty
- if its empty, skip the post
- all items in the list should have a `x` mark after the id
- clicking it should mark the account id for deletion from account roles table
- when user clicks `Install`, all the accounts which have been marked for deletion should be removed from account roles table

stephanielingwood:
complete. closing.

PR #634, opened 2017-05-16
Show superUser management panel

stephanielingwood:
https://github.com/Shippable/admiral/issues/589

- Add superUsers functionality to the addons panel

superusers

### DIFF
--- a/static/scripts/dashboard/dashboardCtrl.js
+++ b/static/scripts/dashboard/dashboardCtrl.js
@@ -363,6 +363,12 @@
           isEnabled: false
         }
       },
+      superUsers: {
+        addingSuperUser: false,
+        deletingSuperUser: false,
+        newSuperUser: '',
+        superUsers: []
+      },
       installingAddons: false,
       systemSettings: {
         db: {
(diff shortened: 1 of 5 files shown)


static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -2531,14 +2634,33 @@
...
+          return next();
         }
       );
     }
 
+    function addSuperUser() {
+      $scope.vm.superUsers.addedSuperUsers.push(
+        $scope.vm.superUsers.newSuperUser
+      );
+      $scope.vm.superUsers.superUsers.push(
```
a-murphy: What happens if you add and remove a superUser before clicking install?

static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -2509,6 +2530,88 @@
...
+          admiralApiAdapter.postSuperUser(newSuperUser,
+            function (err) {
+              if (err)
+                return done(err);
+              return done();
+            }
+          );
+        },
+        function (err) {
+          $scope.vm.superUsers.addedSuperUsers = [];
```
a-murphy: Should this be set to an empty array on an error, or should we give the user a chance to correct it?

Issue #632, opened 2017-05-16
8413 Update CLI prompts


a-murphy:
Shippable/pm#8413

From #591:
  - Once admiral cli is complete, the user prompt should show `Go to <ip address> to access admin panel` instead of `Go to <ip address> to access admiral`
  - Collect all answers to prompts and confirm once (after displaying all of the answers in a table-like format) instead of confirming each input.

PR #639, opened 2017-05-16
#632; updates CLI prompts.

a-murphy:
#632 

Groups all of the confirmations, except the SSH key to install a database, into one prompt, with a summary of the values to be confirmed.

### DIFF
--- a/common/scripts/lib/_parseArgs.sh
+++ b/common/scripts/lib/_parseArgs.sh
@@ -18,7 +18,7 @@ __validate_runtime() {
 
   export OS_TYPE=docker
 
-  ################## check release   #############################
+  ################## set release   ###############################
   local head_sha=$(git symbolic-ref HEAD --short -q)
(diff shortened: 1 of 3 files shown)


common/scripts/lib/_parseArgs.sh
```diff
@@ -95,28 +132,130 @@ __validate_runtime() {
...
     __set_public_image_registry
+    setPublicImageRegistry=true
   else
     __process_msg "PUBLIC_IMAGE_REGISTRY already set, skipping"
   fi
 
-  source $ADMIRAL_ENV
+  ################## confirm values #######################################
+  if $setAccessKey || $setSecretKey || $setAdmiralIP || $setDBIP || $setDBPort || $setDBPassword || $setPublicImageRegistry ; then
+    __process_success "Please confirm that the following values are correct:"
```
stephanielingwood: These values are easy to set now, but hard to change later! Please confirm that they are correct:

Issue #648, opened 2017-05-17
8413 Pull service images before api


stephanielingwood:
https://github.com/Shippable/pm/issues/8413

- The images for postgres, vault, rabbit, gitlab, and redis should be pulled before api.

trriplejay:
complete.

PR #654, opened 2017-05-17
change the pull order

trriplejay:
#648

### DIFF
--- a/common/scripts/lib/_helpers.sh
+++ b/common/scripts/lib/_helpers.sh
@@ -115,18 +115,19 @@ __registry_login() {
 
 __pull_images() {
   __process_marker "Pulling latest service images"
-  __process_msg "Registry: $PRIVATE_IMAGE_REGISTRY"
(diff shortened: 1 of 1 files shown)


common/scripts/lib/_helpers.sh
```diff
@@ -115,18 +115,18 @@ __registry_login() {
...
-    image="$PRIVATE_IMAGE_REGISTRY/$image:$RELEASE"
+  for image in "${PUBLIC_REGISTRY_IMAGES[@]}"; do
+    image="$PUBLIC_IMAGE_REGISTRY/$image:$RELEASE"
     __process_msg "Pulling $image"
     sudo docker pull $image
   done
 
-  for image in "${PUBLIC_REGISTRY_IMAGES[@]}"; do
-    image="$PUBLIC_IMAGE_REGISTRY/$image:$RELEASE"
+  __process_msg "Registry: $PRIVATE_IMAGE_REGISTRY"
```
a-murphy: Do we want a `Registry:` log for the public images, too?  It might help organize the logs.

Issue #694, opened 2017-05-23
8482.1  add cleanup to upgrade workflow


ric03uec:
- https://github.com/Shippable/pm/issues/8482
- add master and worker cleanup routes to upgrade workflow

PR #697, opened 2017-05-24
fixing upgrade workflow

ric03uec:
- https://github.com/Shippable/admiral/issues/694

### DIFF
--- a/common/scripts/lib/_admiralAdapter.sh
+++ b/common/scripts/lib/_admiralAdapter.sh
@@ -141,6 +141,11 @@ _shippable_get_systemSettings() {
   __shippable_get $system_settings_get_endpoint
 }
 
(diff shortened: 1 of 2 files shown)


common/scripts/lib/_admiralAdapter.sh
```diff
@@ -183,6 +188,17 @@ _shippable_post_services() {
   __shippable_post $db_post_endpoint "$body"
 }
 
+_shippable_post_master_cleanup() {
+  local master_post_endpoint="master/cleanup"
+  __shippable_post $master_post_endpoint
```
a-murphy: Does this work without a body?

ric03uec: ya. this doesnt need body

Issue #706, opened 2017-05-24
8491.1 Add checkboxes for SCM to the auth section of the install panel


stephanielingwood:
https://github.com/Shippable/pm/issues/8491

Add checkboxes for SCM providers to the auth section of the the install panel. If an auth provider is selected, also enable the SCM provider (and show the SCM box as checked and disabled). If the SCM box is checked, only enable the SCM provider. Still require the user to have at least one auth provider.

stephanielingwood:
complete. closing.

PR #717, opened 2017-05-24
Allow providers to be enabled for SCM only

stephanielingwood:
https://github.com/Shippable/admiral/issues/706

- When a provider is selected for auth, the scm is enabled by default (auth won't work well without it)
- A provider can be selected for scm only; this will only enable the provider's scm masterInt. It will not create an auth systemIntegration.

### DIFF
--- a/static/scripts/dashboard/dashboardCtrl.js
+++ b/static/scripts/dashboard/dashboardCtrl.js
@@ -100,7 +100,6 @@
           bitbucketKeys: {
             isEnabled: false,
             masterName: 'bitbucketKeys',
-            scmMasterName: 'bitbucket',
             data: {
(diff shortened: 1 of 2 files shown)


static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -785,11 +797,13 @@
       if (!$scope.vm.dbInitialized) return next();
       if (!$scope.vm.systemSettings.secrets.isInitialized) return next();
 
+      var installFormNonSystemInts = ['defaultSystemMachineImage',
+        'coreServices', 'scm', 'systemSettings'];
```
a-murphy: Why was `coreServices` added?

Issue #742, opened 2017-05-31
8510 skip rabbitmq post if it is already initialized


ric03uec:
- https://github.com/Shippable/pm/issues/8510

PR #743, opened 2017-05-31
skipping rabbitmq update if already initialized

ric03uec:
- https://github.com/Shippable/admiral/issues/742

### DIFF
--- a/static/scripts/dashboard/dashboardCtrl.js
+++ b/static/scripts/dashboard/dashboardCtrl.js
@@ -1398,7 +1398,7 @@
 
 
       async.series([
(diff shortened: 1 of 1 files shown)


static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -1418,7 +1419,26 @@
       );
     }
 
+    function getMsgSystemIntegration(update, next) {
```
a-murphy: Why can't you use `isInitialized` from `systemSettings`?  This doesn't accurately determine if it's initialized; it may have failed.

Issue #753, opened 2017-05-31
8512.2 update init script


ric03uec:
- https://github.com/Shippable/pm/issues/8512
- update `admiral.sh` script to 
    - check the credentials in admiral.env
    - if not, then get the instance ID from ec2-instance-data 
    - set the instance id as login token
    - if db password is not set, set the password to instance data

PR #760, opened 2017-06-01
updating admiral init file

ric03uec:
- https://github.com/Shippable/admiral/issues/753

### DIFF
--- a/common/scripts/admiralInit.sh
+++ b/common/scripts/admiralInit.sh
@@ -1,15 +1,95 @@
 #!/bin/bash -e
 
-readonly LOGS_FILE="/var/log/admiral_init.log"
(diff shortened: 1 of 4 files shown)


common/scripts/admiralInit.sh
```diff
@@ -1,15 +1,95 @@
...
+    exit 1
+  else
+    echo "Instance UUID: $uuid"
+    export UUID=$uuid
+  fi
+}
+
+__set_login_token() {
+  echo "Setting login token"
+  if [ -z "$LOGIN_TOKEN" ] || [ "$LOGIN_TOKEN" == "" ]; then
```
a-murphy: Did you source admiral.env?  Or when will these be non-empty?

PR #764, opened 2017-06-02
adds projectId to views table

ChaitanyaVootla:
fixes: https://github.com/Shippable/base/issues/1349


migrations/migrations.sql
```diff
@@ -655,6 +655,11 @@ do $$
       alter table "views" add column "subscriptionId" varchar(24);
     end if;
 
+    -- Add new coloumn  projectId in views table
```
shrivara: column

Issue #787, opened 2017-06-20
DEV 8539 Display callback URLs for SCM providers on the Admiral U


Riturajcse:
https://github.com/Shippable/pm/issues/8539

Riturajcse:
This is done.

Riturajcse:
Reopening this to correct the urls.

Riturajcse:
Verified in rc.

bitbucket: https://rcapp.shippable.com/auth/bitbucket/58a42affc8edf01000625295/identify
bitbucketServer: https://rcapp.shippable.com/auth/bitbucketServer/58d3a3965601c40600c0055b/identify
github: https://rcapp.shippable.com/auth/github/58a42b0007c7cf1300b9a828/identify

PR #788, opened 2017-06-20
displays callback URLs for SCM providers on the Admiral UI

Riturajcse:
https://github.com/Shippable/admiral/issues/787
admiralsystemints


shrivara:
Will this be shown if we refresh the page?

Riturajcse:
@shrivara Yes, It will be there on refreshing the page.

PR #806, opened 2017-06-20
adds runMode and admiral IP checks for systemNode init

Bharath92:
closes https://github.com/Shippable/admiral/issues/794

### DIFF
--- a/api/systemNodes/initialize.js
+++ b/api/systemNodes/initialize.js
@@ -3,14 +3,66 @@
 var self = initialize;
 module.exports = self;
 
+var async = require('async');
+
+var APIAdapter = require('../../common/APIAdapter.js');
+
 function initialize(req, res) {
   var bag = {
     reqBody: req.body,
-    resBody: {}
+    resBody: {},
+    apiAdapter: new APIAdapter(req.headers.authorization.split(' ')[1])
   };
 
   bag.who = util.format('systemNodes|%s', self.name);
   logger.info(bag.who, 'Starting');
 
-  sendJSONResponse(res, bag.resBody);
+  async.series([
+      _checkInputParams.bind(null, bag),
+      _getSystemSettings.bind(null, bag)
+    ],
+    function (err) {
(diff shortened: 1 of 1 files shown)


api/systemNodes/initialize.js
```diff
@@ -3,14 +3,66 @@
...
+
+  bag.apiAdapter.getSystemSettings('',
+    function (err, systemSettings) {
+      if (err)
+        return next(
+          new ActErr(who, ActErr.OperationFailed,
+            'Failed to get system settings : ' + util.inspect(err))
+        );
+
+      if (systemSettings.runMode !== 'dev')
```
ric03uec: run mode is available in global config. no need to call db to get it

Bharath92: isn't that the runMode of admiral? global.config.runMode of admiral doesn't get updated if we update the runMode using the installPanel of admiral. so, fetching it from the db.

api/systemNodes/initialize.js
```diff
@@ -3,14 +3,66 @@
...
+      sendJSONResponse(res, bag.resBody);
+    }
+  );
+}
+
+function _checkInputParams(bag, next) {
+  var who = bag.who + '|' + _checkInputParams.name;
+  logger.verbose(who, 'Inside');
+
+  if (global.config.admiralIP !== 'localhost' &&
```
ric03uec: the route should do what it is intended for i.e. initialize a system node. the separation of concerns here is like this
- api provides a functionality
- client(ui/script) decides whether to use it or not
this means that this api will **always** create a system node no matter what. we will update the client to call it conditionally.

api/systemNodes/initialize.js
ric03uec: ok. actually, we dont even need to do the runmode check. please see my next comment. this route behavior(and its functionality) should not change because of runmode.

Bharath92:
closing this, since this is not required, after discussion with @ric03uec

Issue #815, opened 2017-06-22
DEV 8543.3 Add a library to generate uuid in bower.json and gruntfile


Bharath92:
https://github.com/Shippable/pm/issues/8543

Bharath92:
closing this as it is not required

PR #827, opened 2017-06-22
adds pure-uuid package to get uuid

Bharath92:
https://github.com/Shippable/admiral/issues/815

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -52,7 +52,8 @@ module.exports = function (grunt) {
           'static/bower_components/angular-moment/angular-moment.min.js',
           'static/bower_components/angular-sanitize/angular-sanitize.min.js',
           'static/bower_components/angular-cookies/angular-cookies.min.js',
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -52,7 +52,8 @@ module.exports = function (grunt) {
           'static/bower_components/angular-moment/angular-moment.min.js',
           'static/bower_components/angular-sanitize/angular-sanitize.min.js',
           'static/bower_components/angular-cookies/angular-cookies.min.js',
-          'static/bower_components/angular-ui-router/release/angular-ui-router.min.js'
+          'static/bower_components/angular-ui-router/release/angular-ui-router.min.js',
+          'static/bower_components/pure-uuid/uuid.min.js'
```
ric03uec: why do we need an additional package to do this. just use the current unix timestamp and that should work

Bharath92:
closing this, as it is not required

Issue #862, opened 2017-06-27
8552.2 Add isDebug boolean column to runs


shrivara:
https://github.com/Shippable/pm/issues/8552

PR #864, opened 2017-06-27
Add isDebug BOOLEAN column to runs table

shrivara:
https://github.com/Shippable/admiral/issues/862


shrivara:
I've updated the pull request to remove the default. On investigating we found that setting a default while creating the column sets it for all the rows, which could be a problem in production. I've remove the default for now.

Issue #911, opened 2017-07-05
DEV 8587.9 Reduce the time for which Admiral checks whether Secrets service is up from 60 to 30 seconds, if Admiral is installing Secrets on a new machine.


Bharath92:
https://github.com/Shippable/pm/issues/8587

Bharath92:
this is done, closing

PR #925, opened 2017-07-06
adds check if vault is reachable

Bharath92:
https://github.com/Shippable/admiral/issues/911


common/scripts/Ubuntu_14.04/installVault.sh
```diff
@@ -53,9 +53,17 @@ check_vault() {
 }
 
 main() {
+  check_vault=""
   {
-    type vault &> /dev/null && echo "Vault already installed, skipping" && return
-  }
+    type vault &> /dev/null
```
ric03uec: why cant we use the service command here too? we're running vault using upstart so it should be in the service list

Bharath92: i dont think we run vault & gitlab as upstart. `service --status-all 2>&1 | grep vault` doesn't return anything.

Issue #933, opened 2017-07-07
MVP 1 | Update dashboardCtrl.js to set the value of API/WWW URL to swarm worker's IP if a single swarm worker is present.


himanshu0503:
https://github.com/Shippable/pm/issues/8600

PR #935, opened 2017-07-07
Updates files to api/www url to the first workers ip

himanshu0503:
Closes https://github.com/Shippable/admiral/issues/933


static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -1501,6 +1505,20 @@
       );
     }
 
+    function updateInstallForm(workers, next) {
```
shrivara: Lets move this to just after `updateInitializeForm`.

Issue #931, opened 2017-07-06
8598.1 - start rSync via genexec


trriplejay:
Shippable/pm#8598

update admiral so that it starts the rSync service using genexec and the appropriate options instead of stepExec. make sure it works with install as well as upgrade, and any UI actions that restart services, or re-install should always bring it up from genExec

trriplejay:
verified rSync running on genexec in RC

PR #937, opened 2017-07-07
use genexec for rSync

trriplejay:
#931

hardcoding `shipimg` registry. hopefully we can move it to drydock to help with the admiral flow.

also had to make a new api route to get systemCodes so rather than hard code the new 'service' node type code, which genexec needs.

when services are started, the microConfig file has a list of names that belong to genexec (so far just rSync).  It will use different options on the docker run command for genexec images.

testing by running locally and runnign some rSync jobs

### DIFF
--- a/api/services/microConfig.js
+++ b/api/services/microConfig.js
@@ -31,6 +31,7 @@ function microConfig(params, callback) {
       _getServiceUserToken.bind(null, bag),
       _getAPISystemIntegration.bind(null, bag),
       _getMsgSystemIntegration.bind(null, bag),
+      _getSystemCodes.bind(null, bag),
       _generateImage.bind(null, bag),
       _generateEnvs.bind(null, bag),
       _generateMounts.bind(null, bag),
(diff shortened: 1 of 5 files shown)


api/services/microConfig.js
```diff
@@ -190,14 +224,22 @@ function _generateEnvs(bag, next) {
...
+  } else if (bag.component === 'irc') {
     envs = util.format('%s -e %s=%s',
       envs, 'IRC_BOT_NICK', 'shippable');
+  } else if (bag.component === 'genExec') {
+    envs = util.format('%s -e %s=%s',
+      envs, 'SHIPPABLE_AMQP_URL', amqpUrlRoot);
+    envs = util.format('%s -e %s=%s',
+      envs, 'NODE_TYPE_CODE', bag.serviceNodeCode.code);
+    envs = util.format('%s -e %s=%s',
+      envs, 'LISTEN_QUEUE', 'steps.rSync');
```
a-murphy: This won't always be `steps.rSync`.

api/systemCodes/get.js
```diff
@@ -0,0 +1,66 @@
...
+    resBody: {}
+  };
+
+  bag.who = util.format('systemCodes|%s', self.name);
+  logger.info(bag.who, 'Starting');
+
+  async.series([
+    _checkInputParams.bind(null, bag),
+    _get.bind(null, bag)
+  ],
```
a-murphy: Indentation.

Issue #938, opened 2017-07-10
FIx enable dynamic node UI issues


Riturajcse:
https://github.com/Shippable/pm/issues/8604

Riturajcse:
This is complete. Closing

Riturajcse:
Reopening this to add local image path.

Riturajcse:
This is complete.

PR #941, opened 2017-07-10
fixes enable dynamic node UI issues

Riturajcse:
https://github.com/Shippable/admiral/issues/938

image

isAvailable tooltip text
image

isDefault tooltip text
image

shippable ami id tooltip text
image


static/scripts/dashboard/dashboard.html
```diff
@@ -1362,6 +1362,11 @@ <h4 class="panel-title">
                             <div class="row">
                               <div class="col-md-offset-2 col-md-2">
                                 isAvailable
+                                  <span>
+                                    <a title="This is available for the user to select form the dropdown in the subscription settings.">
+                                      <img src="https://shots.jotform.com/kade/Screenshots/blue_question_mark.png" height="13px" style="vertical-align: text-top !important"/>
```
rageshkrishna: @Riturajcse Why is this getting an image from some random URL? Can we not just use font awesome?

Issue #943, opened 2017-07-10
MVP 2 | Update dashboard.html to display loadbalancer message if more than one workers is present


himanshu0503:
https://github.com/Shippable/pm/issues/8600

PR #944, opened 2017-07-10
Updates file to display LoadBalanced message

himanshu0503:
Fixes https://github.com/Shippable/admiral/issues/943
image


ric03uec:
@himanshu0503 
- we're using a tooltip to show help messages like in https://github.com/Shippable/admiral/pull/941. lets use the same here
- the message here should be `This field requires a FQDN where Shippable API is available`

Issue #969, opened 2017-07-11
DEV 8617 Create a new route GET /api/passthrough/ami/:id to get images by ami-id


Riturajcse:
https://github.com/Shippable/pm/issues/8617

Riturajcse:
This is complete and verified in rc.

PR #976, opened 2017-07-12
adds passthrough route for getting ami by amiId

Riturajcse:
https://github.com/Shippable/admiral/issues/969

### DIFF
--- a/api/passthrough/getImageByAmiId.js
+++ b/api/passthrough/getImageByAmiId.js
@@ -7,14 +7,17 @@ var async = require('async');
 var _ = require('underscore');
 
 var AWSAdapter = require('../../common/awsAdapter.js');
+var APIAdapter = require('../../common/APIAdapter.js');
 
 function getImageByAmiId(req, res) {
   var bag = {
     reqQuery: req.query,
-    accessToken: req.query.accessToken,
-    secretToken: req.query.secretToken,
+    accessToken: null,
+    secretToken: null,
     region: req.query.region || 'us-east-1',
     amiId: req.params.id,
+    systemIntegration: {},
+    apiAdapter: new APIAdapter(req.headers.authorization.split(' ')[1]),
     resBody: {},
     awsAdapter: null
   };
@@ -24,6 +27,7 @@ function getImageByAmiId(req, res) {
 
   async.series([
     _checkInputParams.bind(null, bag),
+    _getProvisionSystemIntegration.bind(null, bag),
     _initializeAWSAdapter.bind(null, bag),
     _getImageFromAmiId.bind(null, bag)
   ],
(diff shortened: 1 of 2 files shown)


api/passthrough/getImageByAmiId.js
```diff
@@ -0,0 +1,102 @@
...
+  bag.awsAdapter.EC2.getImages(bag.amiId,
+    function (err, images) {
+      if (err)
+        return next(
+          new ActErr(who, ActErr.OperationFailed,
+            'Adapter call failed for EC2.getImages',
+            util.inspect(err))
+        );
+
+      var imgId = _.pluck(images, 'ImageId');
```
shrivara: Will this not be an array of image ids?

shrivara: Also, what happens when no images are returned?

api/passthrough/getImageByAmiId.js
```diff
@@ -0,0 +1,102 @@
...
+
+var async = require('async');
+var _ = require('underscore');
+
+var AWSAdapter = require('../../common/awsAdapter.js');
+
+function getImageByAmiId(req, res) {
+  var bag = {
+    reqQuery: req.query,
+    accessToken: req.query.accessToken,
```
ric03uec: the tokens are already available in admiral.env file. why are they being passed in request query?

Riturajcse: This is the accessToken from the UI that the user enters while enabling dynamic node.

Riturajcse: image

Riturajcse:
@ric03uec As discussed I have tested there wont be any issues in calling `updateProvisionSystemIntegration` before `updateSystemMachineImages` in dashboardCtrl.js so I have updated my api passthrough to get access and secret keys from the systemIntegration itself instead of query params. I will send dashboardCtrl changes along with my next changes. As this is a new route and not being called from anywhere so wont be any problem in merging this before updating the dashboardCtrl.

api/passthrough/getImageByAmiId.js
```diff
@@ -0,0 +1,121 @@
...
+
+      if (!systemIntegrations.length) 
+        return next(
+          new ActErr(who, ActErr.OperationFailed,
+            'No provision systemIntegration found: ')
+        );
+
+      bag.systemIntegration = systemIntegrations[0];
+      if (bag.systemIntegration && bag.systemIntegration.data) {
+        bag.accessToken = bag.systemIntegration.data.accessKey;
```
ric03uec: what happens if the keys are not present? will the initialize step fail?

Riturajcse: yes this will fail with proper errors.

Issue #970, opened 2017-07-11
DEV 8617 Update dashboardCtrl.js to check if ami-id is present in aws before updating/posting systemMachineImage


Riturajcse:
https://github.com/Shippable/pm/issues/8617
- Skip this check if `access/secret` key-pair is not present in the installer.

Riturajcse:
This is complete.
Closing.

Riturajcse:
Closed by mistake.

Riturajcse:
This is complete and verified in rc.

PR #980, opened 2017-07-13
shows single error for invalid amiId in admiral UI

Riturajcse:
https://github.com/Shippable/admiral/issues/970

### DIFF
--- a/api/passthrough/getImageByAmiId.js
+++ b/api/passthrough/getImageByAmiId.js
@@ -107,7 +107,11 @@ function _getImageFromAmiId(bag, next) {
   bag.awsAdapter.EC2.getImages(bag.amiId,
     function (err, images) {
       if (err)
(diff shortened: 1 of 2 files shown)


api/passthrough/getImageByAmiId.js
```diff
@@ -107,11 +107,7 @@ function _getImageFromAmiId(bag, next) {
   bag.awsAdapter.EC2.getImages(bag.amiId,
     function (err, images) {
       if (err)
-        return next(
```
ric03uec: whats the problem with returning the Acterr object?

Riturajcse: With this horn error was coming twice one form api and same from UI, so it was required to remove from any one place.
If we want Acterr we can remove the horn from the UI, should I remove from UI instead of this ?

Issue #987, opened 2017-07-18
8653.3 Update Admiral to show link for Docker Installation and add other docker functions.


nranjanShippable:
https://github.com/Shippable/pm/issues/8653

nranjanShippable:
This is in RC.

PR #990, opened 2017-07-18
Installs docker Version 1.13

nranjanShippable:
https://github.com/Shippable/admiral/issues/987

Tested with installing in AWS with docker version 1.12 and 1.13+.

### DIFF
--- a/admiral.sh
+++ b/admiral.sh
@@ -32,6 +32,7 @@ readonly AWSCLI_VERSION=1.10.63
 readonly API_TIMEOUT=600
 readonly ADMIRAL_PORT=50003
 export LC_ALL=C
+export docker_restart=false
 
 # Installation default values #############################
 ###########################################################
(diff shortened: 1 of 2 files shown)


admiral.sh
```diff
@@ -32,6 +32,7 @@ readonly AWSCLI_VERSION=1.10.63
 readonly API_TIMEOUT=600
 readonly ADMIRAL_PORT=50003
 export LC_ALL=C
+export docker_restart=false
```
ric03uec: why do we need this?

nranjanShippable: When we change docker options, to add shippable docker options. We need to restart docker after that, so this acts as a flag.

nranjanShippable:
@ric03uec Right there is a difference with the existing script that here we are not getting static docker binaries.

admiral.sh
ric03uec: i dont think we need to do this for local installation. once docker boots up, everything else should just work without changes. Setting docker opts and restarting is unnecessary. unless im missing something here?

Issue #991, opened 2017-07-19
DEV 813 MVP 42 Pull genexec image from drydock instead of shipimg org in docker hub.


kanthchandra:
from https://github.com/Shippable/pm/issues/8613

1.    update common/scripts/lib/_helpers.sh LINE 170 to use drydock/genexec
2.   Update api/services/microConfig.js:180: to use drydock/genexec
3.    Update common/scripts/configs/default_system_machine_image.sql.template:8: to say genexec/drydock

PR #992, opened 2017-07-19
Updates files to point genexec image to drydock repo instead of shipimg

himanshu0503:
Fixes https://github.com/Shippable/admiral/issues/991

* For new installations the images is getting set to drydock instead of shipimg.
* For existing installations users will need to manually update the systemMachineImage from the UI.

### DIFF
--- a/api/services/microConfig.js
+++ b/api/services/microConfig.js
@@ -16,6 +16,7 @@ function microConfig(params, callback) {
     config: params.config,
     name: params.name,
     registry: params.registry,
+    publicRegistry: params.publicRegistry,
     envs: '',
     mounts: '',
     runCommand: '',
@@ -177,7 +178,7 @@ function _generateImage(bag, next) {
 
   if (bag.component === 'genExec')
     bag.config.image = util.format('%s/genexec:%s',
-      'drydock', bag.releaseVersion);
+      bag.publicRegistry, bag.releaseVersion);
   else
     bag.config.image = util.format('%s/micro:%s',
       bag.registry, bag.releaseVersion);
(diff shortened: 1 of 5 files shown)


common/scripts/lib/_helpers.sh
```diff
@@ -182,7 +182,7 @@ __pull_images() {
 
   __process_msg "Registry: shipimg"
   
-  image="shipimg/genexec:$RELEASE"
+  image="drydock/genexec:$RELEASE"
```
ric03uec: this value should come from `PUBLIC_IMAGE_REGISTRY` env

api/services/microConfig.js
```diff
@@ -177,7 +177,7 @@ function _generateImage(bag, next) {
 
   if (bag.component === 'genExec')
     bag.config.image = util.format('%s/genexec:%s',
-      'shipimg', bag.releaseVersion);
+      'drydock', bag.releaseVersion);
```
ric03uec: can we not use the env value PUBLIC_IMAGE_REGISTRY here instead of hardcoding `drydock`

common/scripts/configs/default_system_machine_image.sql.template
```diff
@@ -5,7 +5,7 @@ do $$
        "isAvailable", "isDefault", "region", "keyName", "runShImage",  "securityGroup",
        "subnetId", "drydockTag", "drydockFamily", "createdBy", "updatedBy", "createdAt", "updatedAt")
      values ('572c81cb39a5440c0031b61c', 'ami-abcdefgh', 'AWS', 'Stable', 'Stable AMI',
-       true, true, 'us-east-1', 'shippable-beta', 'shipimg/genexec:<%= obj.releaseVersion %>', 'sg-123456',
+       true, true, 'us-east-1', 'shippable-beta', 'drydock/genexec:<%= obj.releaseVersion %>', 'sg-123456',
```
ric03uec: see previous comment

Issue #997, opened 2017-07-20
DEV 8673.1 Users should be able to list all available versions for Admiral


niranjan92:
https://github.com/Shippable/pm/issues/8673

- [x] add list command to Admiral to check all available versions
- [x] update admiral help option to show list command

niranjan92:
closing as this is done

PR #999, opened 2017-07-20
add list option to admiral help

niranjan92:
https://github.com/Shippable/admiral/issues/997

adm2

### DIFF
--- a/common/scripts/lib/_parseArgs.sh
+++ b/common/scripts/lib/_parseArgs.sh
@@ -347,7 +347,7 @@ __print_help() {
 __print_list() {
   printf "\nAvailable Admiral versions are:\n"
   git tag
-  printf "\nTo upgrade to a version use: #./admiral.sh upgrade <version>\n\n"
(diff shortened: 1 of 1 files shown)


common/scripts/lib/_parseArgs.sh
```diff
@@ -339,14 +339,15 @@ __print_help() {
...
+    list            Print available admiral versions
   "
   exit 0
 }
 
 __print_list() {
   printf "\nAvailable Admiral versions are:\n"
   git tag
-  printf "\nTo upgrade to a version use: $./admiral.sh upgrade <version>\n\n"
+  printf "\nTo upgrade to a version use: #./admiral.sh upgrade <version>\n\n"
```
himanshu0503: The command should be `./admiral.sh upgrade <version>`

niranjan92: i had added `#` for root user. will change this this to `$sudo` for consistency

ric03uec: lets keep it `$`. we're not using `#` anywhere so if we make this change, we'll have to do it globally

Issue #1016, opened 2017-08-01
8718.1 DEV Create new PUT /api/db route which allows updating address, port, isProcessing, isFailed, isInstalled, isInitialized fields of db systemSettings


amalfra:
https://github.com/Shippable/pm/issues/8718

PR #1023, opened 2017-08-01
Create PUT /api/db route which allows updating address, port, isProce…

amalfra:
…ssing, isFailed, isInstalled, isInitialized fields of db systemSettings

Fixes https://github.com/Shippable/admiral/issues/1016

### DIFF
--- a/api/db/put.js
+++ b/api/db/put.js
@@ -51,20 +51,20 @@ function _get(bag, next) {
   logger.verbose(who, 'Inside');
 
   configHandler.get(bag.component,
-    function (err, state) {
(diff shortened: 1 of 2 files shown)


api/db/put.js
```diff
@@ -0,0 +1,107 @@
...
+
+  return next();
+}
+
+function _get(bag, next) {
+  var who = bag.who + '|' + _get.name;
+  logger.verbose(who, 'Inside');
+
+  configHandler.get(bag.component,
+    function (err, state) {
```
ric03uec: it should be db right? not `state`

Issue #1034, opened 2017-08-07
MVP 2 | Update dashboardCtrl/html to display ACS masterIntegration


himanshu0503:
Shippable/pm#8740

PR #1035, opened 2017-08-07
Updates files to display ACS masterIntegration

himanshu0503:
Fixes https://github.com/Shippable/admiral/issues/1034
image

### DIFF
--- a/static/scripts/dashboard/dashboard.html
+++ b/static/scripts/dashboard/dashboard.html
@@ -2200,6 +2200,12 @@
                           <div class="col-md-12">
                             <div class="checkbox">
                               <label>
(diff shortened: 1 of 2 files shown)


ric03uec:
is this bringing up correct services and updating right values in DB. we need to mention in the comment on how PR changes are being tested

jatindogra:
@himanshu0503  @ric03uec We don't use Azure Container Service integration anymore. This was created even before pipelines. 
We now have an integration Azure DC/OS for the same.

I don't think so we should add this in the Admiral.

Also, we should open a heap for clean up of ACS.

himanshu0503:
Closing this as per https://github.com/Shippable/heap/issues/683#issuecomment-293175941.
@jatindogra will open a heap for the same.

Issue #1069, opened 2017-08-21
MVP 3 | Update services/get.js to add new services in services.json to the systemSettings.services


himanshu0503:
https://github.com/Shippable/pm/issues/8781

PR #1070, opened 2017-08-21
Updates files to save new services from services.json

himanshu0503:
Fixes https://github.com/Shippable/admiral/issues/1069

### DIFF
--- a/api/services/get.js
+++ b/api/services/get.js
@@ -5,9 +5,6 @@ module.exports = self;
 
 var async = require('async');
 var _ = require('underscore');
-var fs = require('fs');
-var path = require('path');
-
 var configHandler = require('../../common/configHandler.js');
 
 function get(req, res) {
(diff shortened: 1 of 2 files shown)


api/services/get.js
```diff
@@ -16,7 +13,8 @@ function get(req, res) {
     resBody: [],
     initializeDefault: false,
     component: 'services',
-    defaultService: {}
+    defaultService: {},
+    services: require('../../common/scripts/configs/services.json')
```
ric03uec: why are we not using relative path for this?

api/services/get.js
```diff
@@ -5,18 +5,18 @@ module.exports = self;
...
 
 function get(req, res) {
   var bag = {
     reqQuery: req.query,
     resBody: [],
     initializeDefault: false,
     component: 'services',
-    defaultService: {}
+    defaultService: {},
+    services: require(path.join(global.config.scriptsDir,
```
ric03uec: not sure why are we doing this. the previous workflow is doing better error handling and exits gracefully. is there anything extra we're getting from loading script like this?

himanshu0503: @ric03uec : `Require` parses the JSON files by default, I don't think it's adding any value by adding error handling for JSON files.
If we are doing that we are giving developers the option to break the `services.json`.

Issue #1075, opened 2017-08-22
MVP 5 | Update dashboardCtrl.js and services/post to bootup internalAPI


himanshu0503:
Shippable/pm#8781

PR #1076, opened 2017-08-22
Updates files to start internalAPI service

himanshu0503:
Fixes https://github.com/Shippable/admiral/issues/1075
Scenarios Tested Locally

### New Installation
 - Entered internalAPI url in new installation and hit Install (creates an internalAPI service)
 - Hit install without entering any internalAPI url integration. (doesn't create internalAPI service)

### Existing Installation
- Entered internalAPI url and hit `Save` (Updates the serviceConfig in db to set isEnabled: true, no service restarts)
- Now hit Restart (internalAPI is booted up at port 50004)

- Removed  internalAPI url and hit `Save` (Updates the serviceConfig in db to set isEnabled: false, no service restarts)
- Now hit Restart (internalAPI is removed)

### DIFF
--- a/static/scripts/dashboard/dashboardCtrl.js
+++ b/static/scripts/dashboard/dashboardCtrl.js
@@ -46,7 +46,9 @@
       globalServices: [
         'api',
         'mktg',
-        'www'
+        'www',
+        'internalAPI',
+        'consoleAPI'
       ],
       initializeForm: {
         secrets: {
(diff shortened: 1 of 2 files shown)


static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -2009,6 +2011,17 @@
       );
     }
 
+    function deleteInternalAPIService(bag,  next) {
+      admiralApiAdapter.deleteService('internalAPI',
+        {isEnabled: $scope.vm.installForm.internalAPI.url.isEnabled},
```
Bharath92: we can probably add this in body and use it from there

Issue #1110, opened 2017-09-22
Admiral retry - Move retry logic to lib/_helpers.sh for database


niranjan92:
https://github.com/Shippable/pm/issues/8974

niranjan92:
this is done closing

PR #1111, opened 2017-09-22
move retry logic to shell script

niranjan92:
https://github.com/Shippable/admiral/issues/1110

### DIFF
--- a/common/scripts/docker/installDb.sh
+++ b/common/scripts/docker/installDb.sh
@@ -5,7 +5,7 @@ export TIMEOUT=180
 
 __check_db() {
   __process_msg "Checking $COMPONENT container status on: $DB_IP:$DB_PORT"
-  __check_service_connection $DB_IP $DB_PORT $COMPONENT "180"
+  __check_service_connection "$DB_IP" "$DB_PORT" "$COMPONENT" "$TIMEOUT"
 }
 
 __run_db() {
--- a/common/scripts/docker/installDb.sh
+++ b/common/scripts/docker/installDb.sh
@@ -4,29 +4,8 @@ export DB_IMAGE="drydock/postgres:$RELEASE"
 export TIMEOUT=180
 
 __check_db() {
-  __process_msg "Checking database container status on: $DB_IP:$DB_PORT"
-  local interval=3
-  local counter=0
-  local db_booted=false
-
-
-  __process_msg "Waiting 10 seconds for db boot"
-  sleep 10
-
-  while [ $db_booted != true ] && [ $counter -lt $TIMEOUT ]; do
-    if nc -vz $DB_IP $DB_PORT &>/dev/null; then
-      __process_msg "Database found"
-      db_booted=true
-    else
-      __process_msg "Waiting for database to start"
-      let "counter = $counter + $interval"
-      sleep $interval
-    fi
-  done
-  if [ $db_booted = false ]; then
-    __process_msg "Failed to boot database container"
-    exit 1
-  fi
+  __process_msg "Checking $COMPONENT container status on: $DB_IP:$DB_PORT"
+  __check_service_up $DB_IP $DB_PORT $COMPONENT "180"
 }
 
 __run_db() {
--- a/common/scripts/docker/installDb.sh
+++ b/common/scripts/docker/installDb.sh
@@ -5,7 +5,7 @@ export TIMEOUT=180
 
 __check_db() {
   __process_msg "Checking $COMPONENT container status on: $DB_IP:$DB_PORT"
-  __check_service_up $DB_IP $DB_PORT $COMPONENT "180"
+  __check_service_connection $DB_IP $DB_PORT $COMPONENT "180"
 }
 
 __run_db() {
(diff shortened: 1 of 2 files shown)


common/scripts/docker/installDb.sh
```diff
@@ -5,28 +5,7 @@ export TIMEOUT=180
...
-      __process_msg "Waiting for database to start"
-      let "counter = $counter + $interval"
-      sleep $interval
-    fi
-  done
-  if [ $db_booted = false ]; then
-    __process_msg "Failed to boot database container"
-    exit 1
-  fi
+  __check_service_up $DB_IP $DB_PORT "database" "180"
```
ric03uec: why are we hardcoding `database`? the env `COMPONENT` should already be there. we need to first look at the data that's already available before adding hardcoded stuff like this which degrades the system

common/scripts/lib/_helpers.sh
```diff
@@ -506,6 +506,56 @@ __copy_script_remote() {
   copy_cmd_out=$(eval $copy_cmd)
 }
 
+# accepts arguments $host $port $serviceName $timeout
+# $timeout is optional and defaults to 60 seconds if not defined
+__check_service_up() {
```
ric03uec: this should be `__check_service_connection`

common/scripts/lib/_helpers.sh
```diff
@@ -506,6 +506,56 @@ __copy_script_remote() {
...
+    exit 1
+  fi
+  local service=$3
+  if [ -z "$service" ]; then
+    echo "got empty service name in __check_service_up(), exiting..."
+    exit 1
+  fi
+  local interval=3
+  local counter=0
+  local db_booted=false
```
ric03uec: wrong variable name. make it generic.

common/scripts/docker/installDb.sh
niranjan92: had hard coded name because it was being used in database specific script `installDB.sh`. I have updated it to use `$COMPONENT`

common/scripts/docker/installDb.sh
```diff
@@ -4,29 +4,8 @@ export DB_IMAGE="drydock/postgres:$RELEASE"
...
-      let "counter = $counter + $interval"
-      sleep $interval
-    fi
-  done
-  if [ $db_booted = false ]; then
-    __process_msg "Failed to boot database container"
-    exit 1
-  fi
+  __process_msg "Checking $COMPONENT container status on: $DB_IP:$DB_PORT"
+  __check_service_up $DB_IP $DB_PORT $COMPONENT "180"
```
ric03uec: - double quote all the variables being passed around
- set the 180 value at the top in a variable that shoudl be passed in the function call

ric03uec: function name is still incorrect.?? how is this even being tested ? im pretty sure this PR wasnt tested in any way whatsoever

common/scripts/lib/_helpers.sh
```diff
@@ -506,6 +506,56 @@ __copy_script_remote() {
   copy_cmd_out=$(eval $copy_cmd)
 }
 
+# accepts arguments $host $port $serviceName $timeout
+# $timeout is optional and defaults to 60 seconds if not defined
+__check_service_connection() {
+
+  if [ "$#" != "3" ] && [ "$#" != "4" ]; then
+    echo "invalid arguments to __check_service_connection()"
```
ric03uec: why are we not using __process_error and __process_msg here like the rest of the file?

niranjan92: @ric03uec 
tried using that but it was giving error - 
```
47ff0ededb29b920037c3070ef8eb85d305193e3d73109261d2add2023e31c24
|___ Checking db container status on: localhost:5432
/home/niranjan/shippable/admiral/common/scripts/lib/_helpers.sh: line 548: process_msg: command not found
```

should i source `logger.sh` in `helper.sh`?

common/scripts/lib/_helpers.sh
```diff
@@ -6,6 +6,7 @@
 # TODO: break up this file into smaller, logically-grouped
 # files after we add release and re-install features
 
+source "$LIB_DIR/_logger.sh"
```
ric03uec: why is this needed?

niranjan92: removed the source 
able to run installer correctly 
```
# 14:08:44 #######################################
# Booting database
##################################################
|___ Running database container
|___ Running a new database container
29264c9558140e0cd3e9f39bfb22d41a1f29673d4aa74d4a14bac6ecd6290aa7
|___ Checking db container status on: localhost:5432
|___ Waiting for db to start
|___ db found
|___ Database container successfully running
```

ric03uec:
- can you mention the test steps here. i need to understand what scenarios were tried out.
- this logic will be used for all service retries, so we need to make sure all cases are covered here

niranjan92:
@ric03uec  tested the retry function locally for following -
- empty arguments for $host, $port, $serviceName and correct number of arguments. 
- $timeout should default to 60 seconds if not defined
- checked retry logic by stopping database container locally and calling the function. retry function works as expected. 
- changed logic for remaining services and able to initialize admiral.

Issue #1133, opened 2017-09-29
9017 DEV add new field blockedEmails to systemSettings schema


ankul-shippable:
https://github.com/Shippable/pm/issues/9017

ankul-shippable:
rc verified, closing

PR #1134, opened 2017-09-29
add new field blockedEmails to systemSettings schema 

ankul-shippable:
#1133

### DIFF
--- a/migrations/migrations.sql
+++ b/migrations/migrations.sql
@@ -3925,11 +3925,6 @@ do $$
       alter table "accounts" add column "isFlagged" boolean;
     end if;
 
-    -- Add "blockedEmails" to systemSettings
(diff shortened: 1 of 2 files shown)


migrations/migrations.sql
```diff
@@ -3925,6 +3925,11 @@ do $$
       alter table "accounts" add column "isFlagged" boolean;
     end if;
 
+    -- Add "blockedEmails" to systemSettings
+    if not exists (select 1 from information_schema.columns where table_name = 'systemSettings' and column_name = 'blockedEmails') then
+      alter table "systemSettings" add column "blockedEmails" TEXT;
+    end if;
+
```
jatindogra: We don't add system settings in migrations.sql, we have a different file to do that i.e `system_settings.sql.template`

jatindogra: This might break things if we do it like this.

Issue #1159, opened 2017-10-09
9049 DEV Update dashboardCtrl.js to add ssh-key integration to add-on list on admiral dashboard


deepikasl:
https://github.com/Shippable/pm/issues/9049

deepikasl:
This is done. verified on rc.

Issue #1158, opened 2017-10-09
9049 DEV Update dashboardCtrl.js to map sshKey integration with sshKeys sysInt and enable the integration by default


deepikasl:
https://github.com/Shippable/pm/issues/9049

deepikasl:
This is done. verified on rc.

Issue #1157, opened 2017-10-09
9049 DEV Update admiral/baseMigrations/migrate.js to create sshKeys systemIntegration using sshKey masterIntegration


deepikasl:
https://github.com/Shippable/pm/issues/9049

deepikasl:
This is done.

PR #1160, opened 2017-10-09
sshKey addon can be managed in Admiral

deepikasl:
https://github.com/Shippable/admiral/issues/1157,
https://github.com/Shippable/admiral/issues/1158,
https://github.com/Shippable/admiral/issues/1159

Verified locally when `sshKeys` was already using `ssh-key`,
steps followed:
copied state.json to baseMigrations and ran:
```
$ sudo docker run -v /home/ubuntu/admiral:/home/shippable/admiral/ -v /etc/shippable:/etc/shippable drydock/microbase:master sh -c 'cd /home/shippable/admiral/baseMigrations && node --version && cd .. && npm install && cd - && npm install && node migrate.js'
```
In this case, it dint create a new shhKeys sysInt using `sshKey` int

screenshot from 2017-10-10 12 09 51

### DIFF
--- a/baseMigrations/migrate.js
+++ b/baseMigrations/migrate.js
@@ -897,7 +897,7 @@ function _createSSHKeysSystemIntegration(bag, next) {
 
   var systemIntegration = {
     name: 'sshKeys',
-    masterName: 'ssh-key',
+    masterName: 'sshKey',
     data: {
       publicKey: bag.systemConfig.systemNodePublicKey ||
         bag.stateJson.systemSettings.systemNodePublicKey,
--- a/baseMigrations/migrate.js
+++ b/baseMigrations/migrate.js
@@ -72,6 +72,7 @@ function migrate() {
       _updateMasterIntegrationFields.bind(null, bag),
       _createMsgSystemIntegration.bind(null, bag),
       _createRedisSystemIntegration.bind(null, bag),
+      _getSshKeysSystemIntegration.bind(null,bag),
       _createSSHKeysSystemIntegration.bind(null, bag),
       _createAPISystemIntegration.bind(null, bag),
       _createWWWSystemIntegration.bind(null, bag),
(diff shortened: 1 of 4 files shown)


baseMigrations/migrate.js
```diff
@@ -897,7 +897,7 @@ function _createSSHKeysSystemIntegration(bag, next) {
 
   var systemIntegration = {
     name: 'sshKeys',
-    masterName: 'ssh-key',
+    masterName: 'sshKey',
```
rageshkrishna: How will this skip if the `sshKeys` system integration was already created using the `ssh-key` master integration?

baseMigrations/migrate.js
```diff
@@ -905,6 +930,10 @@ function _createSSHKeysSystemIntegration(bag, next) {
         bag.stateJson.systemSettings.systemNodePrivateKey
     }
   };
+  // if sshKeys system integration is already using ssh-key masterInt
```
shrivara: Why skip here? Can't we just skip the function entirely?

static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -2444,7 +2448,7 @@
 
       var bag = {
         name: 'sshKeys',
-        masterName: $scope.vm.installForm.sshKeys['ssh-key'].masterName,
+        masterName: $scope.vm.installForm.sshKeys['sshKey'].masterName,
```
shrivara: I'm not sure about how this works. We're not creating keys if `ssh-key` is already present. So, those keys will not show up anymore right?

baseMigrations/migrate.js
shrivara: Nevermind, I read this wrong.

shrivara:
@deepikasl, could you please split this into multiple PRs for the different issues created? It will be easier to review if the PR is doing only one thing.

PR #1370, opened 2017-11-21
change strictJobs to waitingJobs

ankul-shippable:
https://github.com/Shippable/pm/issues/9525


jatindogra:
@ankul-shippable How are we dropping strictJobs table that is already created in rc-database?

ankul-shippable:
we'll directly remove it as it has not yet deplyed to prod hence not adding it in migrations.sql

Issue #1371, opened 2017-11-28
Add migration to add projectId column in the jobTestReports model and upsert projectId to all existing jobTestReports


mohit5it9:
https://github.com/Shippable/pm/issues/9648

mohit5it9:
migration script is successfully run on rc

PR #1372, opened 2017-11-28
adds projectId to jobTestReports object

mohit5it9:
https://github.com/Shippable/admiral/issues/1371
image

### DIFF
--- a/migrations/migrations.sql
+++ b/migrations/migrations.sql
@@ -4307,7 +4307,7 @@ do $$
     -- Add projectId column to jobTestReports table and populate projectId for existing jobTestReports
     if not exists (select 1 from information_schema.columns where table_name = 'jobTestReports' and column_name = 'projectId') then
       alter table "jobTestReports" add column "projectId" varchar(24);
-      update "jobTestReports" as t set "projectId" = (select "projectId" from jobs where id = t."jobId");
+      update "jobTestReports" SET "projectId" = jobs."projectId" from jobs where "jobTestReports"."jobId" = "jobs".id;
       alter table "jobTestReports" alter column "projectId" set not null;
       create index "jobTestRepProjIdI" on "jobTestReports" using btree("projectId");
     end if;
--- a/migrations/migrations.sql
+++ b/migrations/migrations.sql
@@ -4303,5 +4303,18 @@ do $$
     if not exists (select 1 from information_schema.columns where table_name = 'accountIntegrations' and column_name = 'isInternal') then
       alter table "accountIntegrations" add column "isInternal" BOOLEAN;
     end if;
+
+    -- Add projectId column to jobTestReports table and populate projectId for existing jobTestReports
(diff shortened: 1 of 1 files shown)


migrations/migrations.sql
```diff
@@ -4303,5 +4303,18 @@ do $$
     if not exists (select 1 from information_schema.columns where table_name = 'accountIntegrations' and column_name = 'isInternal') then
       alter table "accountIntegrations" add column "isInternal" BOOLEAN;
     end if;
+
+    -- Add projectId column to jobTestReports table and populate projectId for existing jobTestReports
+    if not exists (select 1 from information_schema.columns where table_name = 'jobTestReports' and column_name = 'projectId') then
+      alter table "jobTestReports" add column "projectId" varchar(24);
+      update "jobTestReports" as t set "projectId" = (select "projectId" from jobs where id = t."jobId");
```
mohit5it9: will it cause the table to lockup?

rageshkrishna: Do not use sub-queries. This can be done with a simple join.

migrations/migrations.sql
```diff
@@ -4303,5 +4303,18 @@ do $$
     if not exists (select 1 from information_schema.columns where table_name = 'accountIntegrations' and column_name = 'isInternal') then
       alter table "accountIntegrations" add column "isInternal" BOOLEAN;
     end if;
+
+    -- Add projectId column to jobTestReports table and populate projectId for existing jobTestReports
+    if not exists (select 1 from information_schema.columns where table_name = 'jobTestReports' and column_name = 'projectId') then
+      alter table "jobTestReports" add column "projectId" varchar(24);
+      update "jobTestReports" SET "projectId" = jobs."projectId" from jobs where "jobTestReports"."jobId" = "jobs".id;
+      alter table "jobTestReports" alter column "projectId" set not null;
```
rageshkrishna: You cannot do this as it would break blue/green deployments.

migrations/migrations.sql
```diff
@@ -4303,5 +4303,17 @@ do $$
     if not exists (select 1 from information_schema.columns where table_name = 'accountIntegrations' and column_name = 'isInternal') then
       alter table "accountIntegrations" add column "isInternal" BOOLEAN;
     end if;
+
+    -- Add projectId column to jobTestReports table and populate projectId for existing jobTestReports
+    if not exists (select 1 from information_schema.columns where table_name = 'jobTestReports' and column_name = 'projectId') then
+      alter table "jobTestReports" add column "projectId" varchar(24);
+      update "jobTestReports" SET "projectId" = jobs."projectId" from jobs where "jobTestReports"."jobId" = "jobs".id;
```
shrivara: Just noticed -- do we need a null check here so this doesn't happen every time?

rageshkrishna: This happens only when the column is being added, right? It can't happen again after that.

mohit5it9: yes it is a one time run, only when the projectId field is not there

shrivara: Yep, missed that.

Issue #1385, opened 2017-12-06
adds release notes describing the changes


mohit5it9:
https://github.com/Shippable/pm/issues/9699

mohit5it9:
pr is merged for release notes

PR #1386, opened 2017-12-06
adds release notes for heap#1767

mohit5it9:
https://github.com/Shippable/admiral/issues/1385

### DIFF
--- a/releaseNotes/master.md
+++ b/releaseNotes/master.md
@@ -11,9 +11,7 @@ ${RES_VER_DATE}
       - if necessary
 
 ## Fixes
-  - **simple bold title**: brief description
-      - actions
-      - required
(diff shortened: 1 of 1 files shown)


releaseNotes/master.md
```diff
@@ -11,9 +11,7 @@ ${RES_VER_DATE}
       - if necessary
 
 ## Fixes
-  - **simple bold title**: brief description
-      - actions
-      - required
+  - **Fixes Permissions Checks while creating or deleting a job's coverage and test reports**: We have fixed the bug where users with incorrect permissions were able to create and delete coverage and test reports for a job. After this fix, only users with collaborator and owner access to a project can create and delete coverage and test reports.
 
 ## Shippable Server
```
mohit5it9: do we also need to add release notes for server, too?

trriplejay: only if there is a specific change that a server admin would want to know about, like a new option in admin panel or a change in installer behavior, something like that.

Issue #1438, opened 2017-12-18
Add migration script to add a new field lastIntercomProjectsScanAt in the systemSettings object


mohit5it9:
https://github.com/Shippable/pm/issues/9769

mohit5it9:
verified on rc db.
![image](https://user-images.githubusercontent.com/11424742/34097390-fde2d802-e3fe-11e7-938f-6a284d58d077.png)

PR #1439, opened 2017-12-18
adds migrations to add lastIntercomProjectsScanAt to systemSettings table

mohit5it9:
https://github.com/Shippable/admiral/issues/1438
image

### DIFF
--- a/migrations/migrations.sql
+++ b/migrations/migrations.sql
@@ -4412,5 +4412,10 @@ do $$
     if not exists (select 1 from information_schema.columns where table_name = 'dailyAggs' and column_name = 'buildJobsCancelled') then
       alter table "dailyAggs" add column "buildJobsCancelled" INTEGER;
     end if;
+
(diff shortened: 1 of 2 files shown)


migrations/migrations.sql
```diff
@@ -4412,5 +4412,10 @@ do $$
     if not exists (select 1 from information_schema.columns where table_name = 'dailyAggs' and column_name = 'buildJobsCancelled') then
       alter table "dailyAggs" add column "buildJobsCancelled" INTEGER;
     end if;
+
+    -- Add lastIntercomProjectsScanAt column to systemSettings table
```
shrivara: I think `systemSettings` changes go here: https://github.com/Shippable/admiral/blob/master/common/scripts/configs/system_settings.sql.template

rageshkrishna: You're right. I didn't notice the filename. Sorry, my bad.

mohit5it9: updated. did not know this

Issue #1464, opened 2017-12-20
make admiral confirmation and installer inputs case insensitive


niranjan92:
https://github.com/Shippable/pm/issues/9808

niranjan92:
closing as this is done already

PR #1465, opened 2017-12-20
admiral confirmation and install params should be case insensitive

niranjan92:
https://github.com/Shippable/admiral/issues/1464

### DIFF
--- a/common/scripts/lib/_helpers.sh
+++ b/common/scripts/lib/_helpers.sh
@@ -312,7 +312,7 @@ __set_admiral_ip() {
   __process_success "Please enter your current IP address. This will be the address at which you access the installer webpage. Type D to set default (127.0.0.1) value."
   read response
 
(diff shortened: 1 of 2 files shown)


common/scripts/lib/_helpers.sh
```diff
@@ -325,7 +325,7 @@ __set_db_ip() {
   __process_success "Please enter the IP address of the database or D to set the default ($db_ip)."
   read response
 
-  if [ "$response" != "D" ]; then
+  if [ "$response" != "D" ] && ["$response" != "d"]; then
```
ric03uec: why is this not being converted to toUpper while other inputs are ?

niranjan92: @ric03uec can we give non - ip inputs for admiral IP? those urls will be converted to uppercase. 
also if i give inputs like `localhost` it gets converted to upper case.

Issue #1468, opened 2017-12-21
DEV 9811.1. update getSystemSettings in dashboardCtrl.js to logout of admiral if token is invalid


niranjan92:
https://github.com/Shippable/pm/issues/9811

niranjan92:
this is verified on RC. 

closing

niranjan92:
![anim](https://user-images.githubusercontent.com/1474302/34256459-754c5520-e67b-11e7-8079-0f1ef14f32ed.gif)

PR #1474, opened 2017-12-21
redirect to login page if token is invalid

niranjan92:
https://github.com/Shippable/admiral/issues/1468

anim

### DIFF
--- a/static/scripts/dashboard/dashboardCtrl.js
+++ b/static/scripts/dashboard/dashboardCtrl.js
@@ -682,8 +682,22 @@
     function getSystemSettings(bag, next) {
       admiralApiAdapter.getSystemSettings(
         function (err, systemSettings) {
-          if (err)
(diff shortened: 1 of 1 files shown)


static/scripts/dashboard/dashboardCtrl.js
```diff
@@ -682,8 +682,22 @@
...
-            return next(err);
+          if (err) {
+            if (err.id === 2000) {
+              admiralApiAdapter.postLogout({},
+                function (logoutErr) {
+                  if (logoutErr)
+                    return next(logoutErr);
+
+                  $state.go('login', $state.params);
+                  return next();
```
Bharath92: is this required? if it is, should it be `return next(err);`?

ric03uec:
why are we seeing the error message. the screen should be exactly like the user is visiting for the first time

Issue #1490, opened 2018-01-03
9829 DEV Update /common/scripts/lib/_helpers.sh to check if the entered IP address is valid or not


deepikasl:
https://github.com/Shippable/pm/issues/9829

deepikasl:
This is done.

PR #1491, opened 2018-01-03
validate IP address while configuring admiral

deepikasl:
https://github.com/Shippable/admiral/issues/1490

### DIFF
--- a/common/scripts/lib/_helpers.sh
+++ b/common/scripts/lib/_helpers.sh
@@ -305,6 +305,12 @@ __set_secret_key() {
   export SECRET_KEY="$response"
 }
 
+_validate_ip() {
+  local ip=$1
+  local rx='([1-9]?[0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])'
+  [[ $ip =~ ^$rx\.$rx\.$rx\.$rx$ ]];
+}
+
 __set_admiral_ip() {
   __process_msg "Setting value of admiral IP address"
   local admiral_ip='127.0.0.1'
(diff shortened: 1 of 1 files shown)


common/scripts/lib/_helpers.sh
```diff
@@ -312,8 +312,15 @@ __set_admiral_ip() {
   __process_success "Please enter your current IP address. This will be the address at which you access the installer webpage. Type D to set default (127.0.0.1) value."
   read response
 
+  local rx='([1-9]?[0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])'
```
ric03uec: lets move this into a separate function called `_validate_ip`. use that function anywhere where validation is required

deepikasl: ok

common/scripts/lib/_helpers.sh
```diff
@@ -305,6 +305,18 @@ __set_secret_key() {
...
 
+_validate_ip() {
+  local ip=$1
+  local step=$2
+  local rx='([1-9]?[0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])'
+  if [[ $ip =~ ^$rx\.$rx\.$rx\.$rx$ ]]; then
+    return
+  else
+    __process_error "Invalid response, please enter valid IP address"
+    $step
```
ric03uec: why is this required?
- this should be a simple function that returns true/false and the caller needs to decide what to with the return value. this implementation is too complex for a simple problem

Issue #1503, opened 2018-01-05
9837 DEV Update common/scripts/lib/_helpers.sh to list the machine ips before assigning admiral_ip


deepikasl:
https://github.com/Shippable/pm/issues/9837

deepikasl:
This is done. Verified by configuring admiral.

PR #1504, opened 2018-01-05
Update common/scripts/lib/_helpers.sh to list the machine ips before assigning admiral_ip

deepikasl:
https://github.com/Shippable/admiral/issues/1503
  
screenshot from 2018-01-05 16 55 22

### DIFF
--- a/common/scripts/lib/_helpers.sh
+++ b/common/scripts/lib/_helpers.sh
@@ -307,7 +307,11 @@ __set_secret_key() {
   local get_ips=$(ip -f inet -4 addr | grep inet | grep -v 127.0.0.1 | awk '{print $2}' | awk -F "/" '{print $1}');
 
   if [[ "$get_ips" ]]; then
+    number_of_ips=($get_ips)
+    __process_msg "${#number_of_ips[@]} IP addresses available for the host"
     export MACHINE_IPS=$get_ips
+  else
+    __process_msg "No IP addresses avaiable"
   fi
 }
 
@@ -320,7 +324,7 @@ _validate_ip() {
 __set_admiral_ip() {
   __process_msg "Setting value of admiral IP address"
 
-  __process_msg "List of your machine IP addresses:"
+  __process_msg "List of your machine IP addresses including default:"
 
   local machine_ips=('127.0.0.1')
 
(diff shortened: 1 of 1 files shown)


common/scripts/lib/_helpers.sh
```diff
@@ -303,6 +303,12 @@ __set_secret_key() {
   __process_success "Please enter the provided installer secret key."
   read response
   export SECRET_KEY="$response"
+
+  local get_ips=$(ip -f inet -4 addr | grep inet | grep -v 127.0.0.1 | awk '{print $2}' | awk -F "/" '{print $1}');
+
+  if [[ "$get_ips" ]]; then
```
ric03uec: add a logger here in both if and else blocks to say whether ips are found or not. 
- if found, then something like `x ip addresses available for the host`
- if not, `no ip addresses avaiable`

common/scripts/lib/_helpers.sh
```diff
@@ -313,20 +319,36 @@ _validate_ip() {
...
+
+  for (( i = 1 ; i < ${#machine_ips[@]} + 1 ; i++ )) do
+    __process_msg "$i - (${machine_ips[$i-1]})"
+  done
+
+  __process_success "Please pick one of the IP addresses from the above list by selecting the index (1, 2, etc) or enter a custom IP"
   read response
 
-  if [ "$response" != "D" ] && [ "$response" != "d" ]; then
+  if [[ $response =~ ^[0-9]+$ ]]; then
```
ric03uec: add a logger after this, something like `1.1.1.1 was selected as admiral ip`

common/scripts/lib/_helpers.sh
```diff
@@ -313,20 +319,36 @@ _validate_ip() {
...
+    if [ ${machine_ips[$response-1]} ]; then
+      export ADMIRAL_IP=${machine_ips[$response-1]}
```
ric03uec: add logger afeter this, `admiral ip is set to ...`

ric03uec: we want more verbosity here to make sure user knows exactly what is happening with each step they're taking. we can later hide some of the redundant stuff but the basic assumption should be that user should be shown everything

deepikasl: ok, i will update it.

Issue #1500, opened 2018-01-05
Drop the hubspotId, hubspotShouldSimulate, hubspotProjectsLastSyncTime, hubspotRSyncLastSyncTime fields from systemSettings table.


mohit5it9:
https://github.com/Shippable/pm/issues/9825

mohit5it9:
verified on rc.
hubspot related fields are removed from systemSettings table

Issue #1499, opened 2018-01-05
baseMigrations/migrate_system_settings.sql.template template no longer populates hubspot related field


mohit5it9:
https://github.com/Shippable/pm/issues/9825

mohit5it9:
verified on rc.
hubspot related fields are removed from systemSettings table

Issue #1498, opened 2018-01-05
Remove populating hubspotId, hubspotShouldSimulate, hubspotProjectsLastSyncTime, hubspotRSyncLastSyncTime fields from baseMigrations/migrate.js


mohit5it9:
https://github.com/Shippable/pm/issues/9825

mohit5it9:
verified on rc.
hubspot related fields are removed from systemSettings table

PR #1508, opened 2018-01-08
removes hubspot related fields from systemSettings table

mohit5it9:
https://github.com/Shippable/admiral/issues/1498
https://github.com/Shippable/admiral/issues/1500
https://github.com/Shippable/admiral/issues/1499

tested on fresh install and on admiral upgrade. Hubspot related fields are getting removed from the system.


rageshkrishna:
It still isn't safe to do this yet, is it? The fields are being removed from the API model only with the next release, so wouldn't dropping the columns break blue/green rollback?

mohit5it9:
yeah we'll merge this PR after release 6.1.2. Marking it do not merge till then

Issue #1479, opened 2017-12-22
[bug] Incorrect system code dynamic__x86_64__Ubuntu__14.04__c4.xlarge 


shrivara:
https://github.com/Shippable/pm/issues/9776

PR #1512, opened 2018-01-08
Fix incorrect systemCode

shrivara:
https://github.com/Shippable/admiral/issues/1479


common/scripts/configs/system_codes.sql
```diff
@@ -477,9 +477,10 @@ do $$
       values (10000, 'dynamic__x86_64__Ubuntu_14.04__c4.large', 'clusterType', '54188262bc4d591ba438d62a', '54188262bc4d591ba438d62a', '2017-12-12', '2017-12-12');
     end if;
 
+    delete from "systemCodes" where code = 10001 and name = 'dynamic__x86_64__Ubuntu__14.04__c4.xlarge';
```
shrivara: There was an extra _ between `x86_64` and `Ubuntu`.

Issue #1514, opened 2018-01-09
DEV 9814.1. update api/admiral/get.js to correctly read password with = sign


niranjan92:
https://github.com/Shippable/pm/issues/9814

niranjan92:
this is verified, closing

PR #1516, opened 2018-01-09
admiral split saved variables on first =

niranjan92:
https://github.com/Shippable/admiral/issues/1514


ric03uec:
can we remove the stack overflow reference from here. the comment is sufficient

niranjan92:
@ric03uec removed comment for Stackoverflow reference

Issue #1537, opened 2018-01-11
DEV 9851.1. update admiral.env.template to add IGNORE_TLS_ERRORS parameter


niranjan92:
https://github.com/Shippable/pm/issues/9851

niranjan92:
this is done, closing

PR #1547, opened 2018-01-11
adds env var IGNORE_TLS_ERRORS to admiral.env

niranjan92:
https://github.com/Shippable/admiral/issues/1537


niranjan92:
not updating `baseMigrations/admiral.env.template` as changes to `envHandler put` should create the field if it doesn't exist.

Issue #1523, opened 2018-01-09
9850 DEV List of urls to be validated: API URL, Dedicated Internal API, Console API URL, State URL, RabbitMQ amqp URL, RabbitMQ Root URL, RabbitMQ Admin URL, Redis URL, Shippable UI URL, BitbucketServer URL, GHE URL, Gitlab URL


deepikasl:
https://github.com/Shippable/pm/issues/9850

deepikasl:
This is done. Verified.

PR #1550, opened 2018-01-11
use regex to validate url instead of using angular

deepikasl:
https://github.com/Shippable/admiral/issues/1523

angular $dirty is not accepting `amqp://shippableRoot:shippable@127.0.0.1:5672/shippable` this kind of url as a valid one which is `amqp msg url`.
Hence using regex: `(http(s)?:\/\/)?([a-zA-Z0-9]+(\.[a-zA-Z0-9]+)+.*)$` to validate the admiral urls

### DIFF
--- a/static/scripts/dashboard/dashboard.html
+++ b/static/scripts/dashboard/dashboard.html
@@ -805,12 +805,13 @@
                             </div>
                             <div class="col-md-8">
                               <input type="url" class="form-control input-sm" name="apiUrl"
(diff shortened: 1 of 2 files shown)


rageshkrishna:
I am not convinced that a change this big was necessary to fix it. The documentation for `input` says you can specify a custom pattern with `ng-pattern`: https://docs.angularjs.org/api/ng/input/input%5Burl%5D

Issue #1538, opened 2018-01-11
DEV 9851. 2. on boot, admiral loads IGNORE_TLS_ERRORS flag and sets global.config.ignoreTLSErrors


niranjan92:
https://github.com/Shippable/pm/issues/9851

niranjan92:
this is done, closing

niranjan92:
opening to fix `boot_admiral.sh` script

niranjan92:
this is done, closing

PR #1551, opened 2018-01-11
admiral should set global.config.ignoreTlsErrors on boot

niranjan92:
https://github.com/Shippable/admiral/issues/1538

### DIFF
--- a/admiral.app.js
+++ b/admiral.app.js
@@ -130,6 +130,11 @@ function _initializeConfig(bag, next) {
   else
     bag.config.runMode = 'production';
 
+  if (bag.env.IGNORE_TLS_ERRORS && bag.env.IGNORE_TLS_ERRORS === 'true')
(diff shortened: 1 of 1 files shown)


admiral.app.js
```diff
@@ -130,6 +130,11 @@ function _initializeConfig(bag, next) {
   else
     bag.config.runMode = 'production';
 
+  if (bag.env.IGNORE_TLS_ERRORS && bag.env.IGNORE_TLS_ERRORS === 'true')
```
ric03uec: why will it be a string and not boolean?

niranjan92: @ric03uec : currently in `envHandler.js put`, the `sed` command wraps value in double quotes. so the environment variable is loaded as a string.

https://github.com/Shippable/admiral/blob/master/common/envHandler.js#L37

PR #1554, opened 2018-01-11
use ng-pattern

deepikasl:
https://github.com/Shippable/admiral/issues/1523


rageshkrishna:
Why do you need custom patterns for everything? I thought the problem was only with the amqp url. Everything else should work with the built-in patterns, shouldn't it?

Issue #1544, opened 2018-01-11
DEV 9851. 8. in envHandler.js adds post route to create env var if not present in admiral.env


niranjan92:
https://github.com/Shippable/pm/issues/9851

niranjan92:
this is done, closing

PR #1559, opened 2018-01-11
adds post route for envHandler.js

niranjan92:
https://github.com/Shippable/admiral/issues/1544

new variable gets appended to `admiral.env` file 
verified using postman

### DIFF
--- a/common/envHandler.js
+++ b/common/envHandler.js
@@ -33,6 +33,36 @@ envHandler.get = function (envName, cb) {
   );
 };
 
+envHandler.post = function (envName, envValue, cb) {
+  var createCmd = util.format('sudo echo \'%s="%s"\' >> %s',
+    envName, envValue, global.config.admiralEnv);
+
+  var exec = spawn('/bin/bash',
+    ['-c', createCmd]
+  );
+
+  exec.stdout.on('data',
+    function (data)  {
+      logger.debug(self.name, data.toString());
+    }
+  );
+
+  exec.stderr.on('data',
+    function (data)  {
+      logger.error(self.name, data.toString());
+    }
+  );
+
+  exec.on('close',
+    function (exitCode)  {
+      if (exitCode)
+        return cb(exitCode, null);
+
+      return cb(null);
+    }
+  );
+};
+
 envHandler.put = function (envName, envValue, cb) {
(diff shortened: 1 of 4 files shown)


common/envHandler.js
```diff
@@ -34,8 +35,60 @@ envHandler.get = function (envName, cb) {
...
+        return cb(err, null);
+
+      return cb(null);
+    }
+  );
+};
+
+// adds env variable ENV_VAR="" to admiral.env file if it doesn't exist
+function _createEnvVar(bag, next) {
+  var createCmd = util.format('sudo grep -qo \'%s\' %s || echo \'%s=""\' >> %s '
```
ric03uec: - we cant do this. you're changing put to basically do an upsert. so use the get method to find the env, if its not there, then add a function to insert it else update it. we cant do everything in one bash script. makes things unreadable and clunky

niranjan92: @ric03uec : ok, i 'll create a new route to `post admiral env` and move the logic there

common/envHandler.js
```diff
@@ -33,6 +33,36 @@ envHandler.get = function (envName, cb) {
   );
 };
 
+envHandler.post = function (envName, envValue, cb) {
+  var createCmd = util.format('sudo echo \'%s="%s"\' >> %s',
```
ric03uec: where is the validation for each of the inputs. its not there in the route handler or here. at least check if the values are not null

ric03uec: why is there no check here. we cannot have a null env even though the value can be null. i dont think we're thinking this through. this sort of code creates more problems than it solves because of lack of resiliency

common/envHandler.js
```diff
@@ -33,6 +33,39 @@ envHandler.get = function (envName, cb) {
   );
 };
 
+envHandler.post = function (envName, envValue, cb) {
+  if (envName === null || envValue === null)
```
ric03uec: what if its undefined and not null? use udnerscore here

niranjan92: @ric03uec updated to add checks using underscore.js. we won't reach this point from API because json parsing itself fails if payload contains `undefined` but added the checks anyway for safety.

PR #1571, opened 2018-01-11
load IGNORE_TLS_ERRORS from admiral.env instead of global variable

niranjan92:
tested locally by enabling and disabling checkbox to ignore TLS errors and checking if env var `NODE_TLS_REJECT_UNAUTHORIZED=0` in container
- also disable TLS for admiral service itself. 

fixes
https://github.com/Shippable/admiral/issues/1540
https://github.com/Shippable/admiral/issues/1541
https://github.com/Shippable/admiral/issues/1542
https://github.com/Shippable/admiral/issues/1543

### DIFF
--- a/admiral.app.js
+++ b/admiral.app.js
@@ -130,10 +130,13 @@ function _initializeConfig(bag, next) {
   else
     bag.config.runMode = 'production';
 
-  if (bag.env.IGNORE_TLS_ERRORS && bag.env.IGNORE_TLS_ERRORS === 'true')
+  if (bag.env.IGNORE_TLS_ERRORS && bag.env.IGNORE_TLS_ERRORS === 'true') {
+    // disable TLS check for admiral itself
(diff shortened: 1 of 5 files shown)


admiral.app.js
```diff
@@ -130,10 +130,13 @@ function _initializeConfig(bag, next) {
   else
     bag.config.runMode = 'production';
 
-  if (bag.env.IGNORE_TLS_ERRORS && bag.env.IGNORE_TLS_ERRORS === 'true')
+  if (bag.env.IGNORE_TLS_ERRORS && bag.env.IGNORE_TLS_ERRORS === 'true') {
+    // disable TLS check for admiral itself
```
ric03uec: why are we doing this?

niranjan92: @ric03uec : admiral service was left out so disabled TLS there also consistency if the IGNORE_TLS_ERRORS flag is set

ric03uec: got it

Issue #1593, opened 2018-01-12
9817.15 user should be able to configure gitlab as onebox from the new UI


ric03uec:
- https://github.com/Shippable/pm/issues/9817

mohit5it9:
this is done.closing

PR #1601, opened 2018-01-12
adds state section for onebox installation

mohit5it9:
https://github.com/Shippable/admiral/issues/1593

image

### DIFF
--- a/static/scripts/dashboard/dashboard2.html
+++ b/static/scripts/dashboard/dashboard2.html
@@ -388,14 +388,14 @@
 
 
 
-  <!-- START State -->
-  <div class="row">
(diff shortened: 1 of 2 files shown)


ric03uec:
can you attach screenshot?

mohit5it9:
@ric03uec 
image

ric03uec:
why is it showing add new/existing nodes when dropdown says this node

mohit5it9:
it's disabled. for all vault, rabbitMQ we have also done like this.
so you want to remove add section if "This node" is selected?

ric03uec:
yes, this needs to be disabled

PR #1605, opened 2018-01-12
make state section consistent with the other section panels

mohit5it9:
**after new node install**
image

**after on this node install**
image

https://github.com/Shippable/admiral/issues/1593


ric03uec:
this needs to be in a modal like others panels

ric03uec:
i dont get it. why is the new node install asking for ip address in the same panel? shouldn't this be in the modal instead?

mohit5it9:
no it's not asking for the IP
it's just displaying it after install.
new node form will come up in the modal only

Issue #1627, opened 2018-01-16
9817.29 user should be able to configure build node settings from the new UI


ric03uec:
- https://github.com/Shippable/pm/issues/9817
- user should be able to configure build node settings from the new UI

priyakirankumar:
![image](https://user-images.githubusercontent.com/16892929/34972440-89636aea-faa7-11e7-81ed-7eaeb3e79604.png)

mohit5it9:
verififed on rc

PR #1680, opened 2018-01-18
adds nodes section in the configs panel

mohit5it9:
https://github.com/Shippable/admiral/issues/1627

image

image


ric03uec:
@mohit5it9 can we use the question mark icon from font-awesome only. thsi will make sure the theme remains consistent

mohit5it9:
@ric03uec updated 
image

Issue #1707, opened 2018-01-22
[bug] issues in infrastructure initialize workflow


ric03uec:
- https://github.com/Shippable/pm/issues/9817
- If you do keep clicking on Initialize button in the disabled state, it disappears. Overall, the disabled state of the button is poorly indicated visually. I did not have a clue that the button was disabled because some checkboxes were not clicked (that need to be clicked after the SSH auth script is run) for some component. We should clearly indicate why the button is disabled with a call to action, since in the new UI the initialization is hidden in a panel and not visible. 

<img width="1168" alt="screen shot 2018-01-18 at 12 01 51 pm" src="https://user-images.githubusercontent.com/26096758/35119273-4aed5f02-fc49-11e7-8858-366821a509da.png">
- show status code colors for initializing, failed, initialized steps
    - use the same status code colors that are being used for Shippable
    - show those code when the component is in processing, success or failed etc
- disable the Initialize button when any component is in initializing phase.
- all action buttons on the infra panel should be disabled when any component is being initialized

amalfra:
rc verified

PR #1713, opened 2018-01-22
show status code colors for initialize panels

amalfra:
https://github.com/Shippable/admiral/issues/1707

peek 2018-01-22 15-54

### DIFF
--- a/static/scripts/dashboard/dashboard2.html
+++ b/static/scripts/dashboard/dashboard2.html
@@ -137,11 +137,11 @@
                 </h3>
                 <div class="portlet-widgets">
                   <span class="divider"></span>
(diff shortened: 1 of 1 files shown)


ric03uec:
i dont udnerstand where these colors are coming from. this is very simple fix
- DO NOT change color of the entire bar
- use labels like 
image
- this should not be any differernt from the status codes we show on the main product

ric03uec:
just change the initialized/failed/initializing text to map with what we already have

amalfra:
@ric03uec updated pr and screenshot. The colours are same as in man app since it's using same css classes. The mapping used is:
* initialized - success status colour
* failed - failed status colour
* initializing - processing status colour

Issue #1638, opened 2018-01-16
9817.5 user should be able to see the list of servers in summary section


ric03uec:
- https://github.com/Shippable/pm/issues/9817
- user should be able to see the list of servers in summary section
- show a table like following

|Nodes| Admiral | Database | Secrets | Messaging | State | Redis | Swarm Master | Swarm Workers |
|---|---|---|---|---|---|---|---|---|
|1.1.1.1| x | | | | | | | |
|1.1.1.2|  |x|x| |x| | | |

- a cross should be shown in front of any component if and only it has `isInitialized` and `isInstalled`=true

ric03uec:
@amalfra eod status on this?
can we make sure that issues are getting updated with status.

amalfra:
ok, will do. The pr was opened yesterday night but was submitted for review today only.

amalfra:
rc verified

PR #1714, opened 2018-01-22
add list of servers in summary section

amalfra:
https://github.com/Shippable/admiral/issues/1638

### first time login screenshot
selection_001

### init recording
peek 2018-01-23 13-00


ric03uec:
- remove the message "a detailed summary of this section" message
- what does the user see on first login. i.e. when nothing is initialized?

ric03uec:
- does this table get automatically refreshed or user needs to do it?
- can we test a scenario where this is a two box installation

Issue #1735, opened 2018-01-24
[bug] sometimes on new admiral dashboard page the left hand side panel is showing the scroller needlessly


mohit5it9:
https://github.com/Shippable/pm/issues/9817

![image](https://user-images.githubusercontent.com/11424742/35323624-26096ce2-0114-11e8-87c6-dfcabc970714.png)

priyakirankumar:
Adds empty space below too.

![image](https://user-images.githubusercontent.com/16892929/35324681-6e98b79e-0117-11e8-8d9f-5f221d5b0dc9.png)

priyakirankumar:
Verified on RC.
![image](https://user-images.githubusercontent.com/16892929/35385348-00f59b58-01ee-11e8-9419-7916ec760195.png)

PR #1743, opened 2018-01-25
Removed the class as side panel is showing the scroller needlessly

priyakirankumar:
https://github.com/Shippable/admiral/issues/1735


priyakirankumar:
image

Issue #1800, opened 2018-02-06
Update common/scripts/boot_admiral.sh to copy proxy envs onto the Admiral container if found


shrivara:
https://github.com/Shippable/pm/issues/10051

PR #1801, opened 2018-02-06
Bring up admiral with proxy envs if available

shrivara:
https://github.com/Shippable/admiral/issues/1800


common/scripts/lib/_helpers.sh
```diff
@@ -120,9 +120,10 @@ __check_dependencies() {
     rm installDockerScript.sh
   fi
 
+  #
```
shrivara: Just cleaning up comments from a previous issue.

Issue #1817, opened 2018-02-07
Add showBetaSPOGView to systemCodes


ChaitanyaVootla:
https://github.com/Shippable/pm/issues/9821

PR #1818, opened 2018-02-07
adds showBetaSPOGView to systemCodes

ChaitanyaVootla:
https://github.com/Shippable/admiral/issues/1817


migrations/migrations.sql
```diff
@@ -779,6 +779,12 @@ do $$
       alter table "jobs" add constraint "jobs_versionId_fkey" foreign key ("versionId") references "versions"(id) on update restrict on delete restrict;
     end if;
 
+    -- add showBetaSPOGView viewObjectType to systemCodes
```
shrivara: We should do this in common/scripts/configs/system_codes.sql.

Issue #1816, opened 2018-02-07
Set the value for NO_VERIFY_SSL to true if --no-verify-ssl is passed.


shrivara:
https://github.com/Shippable/pm/issues/10055

PR #1819, opened 2018-02-07
Set  to true if --no-verify-ssl is passed

shrivara:
https://github.com/Shippable/admiral/issues/1816

### DIFF
--- a/common/scripts/lib/_parseArgs.sh
+++ b/common/scripts/lib/_parseArgs.sh
@@ -51,7 +51,16 @@ __validate_runtime() {
   fi
 
   if [ -z "$NO_VERIFY_SSL" ]; then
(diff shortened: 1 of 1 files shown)


common/scripts/lib/_parseArgs.sh
```diff
@@ -300,6 +310,27 @@ __parse_args_install() {
   fi
 }
 
+__parse_args_upgrade() {
```
shrivara: This function is similar to __parse_args_install above.

common/scripts/lib/_parseArgs.sh
```diff
@@ -464,8 +505,9 @@ __parse_args() {
         __checkout_tag
         ;;
       upgrade)
```
shrivara: This is now similar to the way install behaves.

Issue #1823, opened 2018-02-07
Modify common/scripts/docker/startFakeAPI.sh to pass --no-verify-ssl to ecr's login command if NO_VERIFY_SSL is true


shrivara:
https://github.com/Shippable/pm/issues/10055

PR #1824, opened 2018-02-07
Login using --no-verify-ssl when booting fake API

shrivara:
https://github.com/Shippable/admiral/issues/1823

### DIFF
--- a/api/db/post.js
+++ b/api/db/post.js
@@ -27,7 +27,8 @@ function post(req, res) {
     setServiceUserToken: false,
     accessKeyEnv: 'ACCESS_KEY',
     secretKeyEnv: 'SECRET_KEY',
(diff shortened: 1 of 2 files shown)


api/db/post.js
```diff
@@ -825,6 +847,18 @@ function _runScript(seriesBag, next) {
   var who = seriesBag.who + '|' + _runScript.name;
   logger.verbose(who, 'Inside');
 
+  /* jshint camelcase:false */
+  seriesBag.scriptEnvs = seriesBag.scriptEnvs || {};
```
shrivara: spawn does not copy over these envs from the enviroment. I had missed this in the initial design.

shrivara: I'll add this to all the spawns in a subsequent PR.

Issue #1825, opened 2018-02-07
Modify common/scripts/installWorker.sh to pass --no-verify-ssl to ecr's login command if NO_VERIFY_SSL is true


shrivara:
Shippable/pm#10055

PR #1826, opened 2018-02-07
Use NO_VERIFY_SSL when initializing workers

shrivara:
https://github.com/Shippable/admiral/issues/1825

### DIFF
--- a/common/scripts/Ubuntu_14.04/installWorker.sh
+++ b/common/scripts/Ubuntu_14.04/installWorker.sh
@@ -57,7 +57,11 @@ __initialize_worker() {
 __registry_login() {
   echo "Logging into registry"
   mkdir -p ~/.aws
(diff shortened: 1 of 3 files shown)


common/scripts/Ubuntu_14.04/installWorker.sh
```diff
@@ -57,7 +57,11 @@ __initialize_worker() {
 __registry_login() {
   echo "Logging into registry"
   mkdir -p ~/.aws
-  local docker_login_cmd=$(aws ecr --region us-east-1 get-login)
+  if [ "$NO_VERIFY_SSL" == true ]; then
```
shrivara: This will gets called by `common/scripts/installWorker.sh` below. I'm not sure why login is required twice. I have kept the existing behavior as-is.

Issue #1839, opened 2018-02-08
Ensure that proxy config is correct on the Database machine before installing any dependencies in common/scripts/installDb.sh


shrivara:
https://github.com/Shippable/pm/issues/10049

shrivara:
>

PR #1840, opened 2018-02-08
Configure proxy on database machine

shrivara:
https://github.com/Shippable/admiral/issues/1839

### DIFF
--- a/common/scripts/installDb.sh
+++ b/common/scripts/installDb.sh
@@ -85,6 +85,15 @@ __install_db() {
   else
     __check_connection "$DB_IP"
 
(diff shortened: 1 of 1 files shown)


common/scripts/installDb.sh
```diff
@@ -85,6 +85,15 @@ __install_db() {
   else
     __check_connection "$DB_IP"
 
+    local proxy_script_name="configureProxy.sh"
```
shrivara: Note: This will only configure proxy when the env `DB_INSTALLED` is false and we are not installing on the same machine as the of Admiral (which brings up a Docker container instead).

Issue #1875, opened 2018-02-19
Update AWSCLI to 1.11.91. Update aws ecr get-login command to include --no-include-email flag.


Bharath92:
https://github.com/Shippable/pm/issues/10115

Bharath92:
this is done, closing

PR #1876, opened 2018-02-19
install AWS CLI 1.11.91

Bharath92:
https://github.com/Shippable/admiral/issues/1875


common/scripts/lib/_helpers.sh
```diff
@@ -201,9 +201,16 @@ __check_dependencies() {
...
   if type aws &> /dev/null && true; then
-    __process_msg "'awscli' already installed"
-  else
+    local current_aws_cli_version=$(pip show awscli | awk '/^Version: / { print $2 }')
+    if __version_gte $AWSCLI_VERSION $current_aws_cli_version; then
+      install_awscli=false
+      __process_msg "'awscli' already installed"
+    fi
+  fi
+  if $install_awscli; then
```
ric03uec: can we make this consistent with rest of the syntax

Bharath92:
closing this PR after discussion with @ric03uec AWS CLI, docker versions will not be common across all the OS/architectures. So, we will not be making any changes to U14 scripts

Issue #1884, opened 2018-02-19
10128 DEV Add migration query to create new columns prJobsCount, prJobsSuccessful, prJobsFailed, prJobsTimedOut, prJobsCancelled, prJobsSkipped, prJobsUnstable to projectDailyAggs


deepikasl:
https://github.com/Shippable/pm/issues/10128

deepikasl:
Verified in RC

PR #1885, opened 2018-02-19
Add migration query to create new columns prJobsCount, prJobsSuccessful, prJobsFailed, prJobsTimedOut, prJobsCancelled, prJobsSkipped, prJobsUnstable to projectDailyAggs

deepikasl:
https://github.com/Shippable/admiral/issues/1884

screenshot from 2018-02-20 09 57 39


rageshkrishna:
You cannot make it a not null column. I don't think there's any reason for this to have a default value, either.

Issue #1895, opened 2018-02-20
Complete admiral/common/scripts/x86_64/Ubuntu_14.04/remote/installRedis.sh


scriptnull:
https://github.com/Shippable/pm/issues/10143

scriptnull:
verified by running the script manually on a Ubuntu 14.04 node.

```
ubuntu@ip-172-31-41-167:~/admiral/common/scripts/x86_64/Ubuntu_14.04/remote$ sudo ./installRedis.sh 
installing redis
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following extra packages will be installed:
  libjemalloc1 redis-tools
The following NEW packages will be installed:
  libjemalloc1 redis-server redis-tools
0 upgraded, 3 newly installed, 0 to remove and 0 not upgraded.
Need to get 410 kB of archives.
After this operation, 1,272 kB of additional disk space will be used.
Get:1 http://us-west-2.ec2.archive.ubuntu.com/ubuntu/ trusty/universe libjemalloc1 amd64 3.5.1-2 [76.8 kB]
Get:2 http://us-west-2.ec2.archive.ubuntu.com/ubuntu/ trusty/universe redis-tools amd64 2:2.8.4-2 [65.7 kB]
Get:3 http://us-west-2.ec2.archive.ubuntu.com/ubuntu/ trusty/universe redis-server amd64 2:2.8.4-2 [267 kB]
Fetched 410 kB in 0s (14.7 MB/s)  
Selecting previously unselected package libjemalloc1.
(Reading database ... 52043 files and directories currently installed.)
Preparing to unpack .../libjemalloc1_3.5.1-2_amd64.deb ...
Unpacking libjemalloc1 (3.5.1-2) ...
Selecting previously unselected package redis-tools.
Preparing to unpack .../redis-tools_2%3a2.8.4-2_amd64.deb ...
Unpacking redis-tools (2:2.8.4-2) ...
Selecting previously unselected package redis-server.
Preparing to unpack .../redis-server_2%3a2.8.4-2_amd64.deb ...
Unpacking redis-server (2:2.8.4-2) ...
Processing triggers for man-db (2.6.7.1-1ubuntu1) ...
Processing triggers for ureadahead (0.100.0-16) ...
Setting up libjemalloc1 (3.5.1-2) ...
Setting up redis-tools (2:2.8.4-2) ...
Setting up redis-server (2:2.8.4-2) ...
Starting redis-server: redis-server.
[...]

PR #1896, opened 2018-02-20
Adds redis install script for ubuntu 14.04 

scriptnull:
https://github.com/Shippable/admiral/issues/1895

Verified by directly running the script and trying out `INFO`, `SET`, `GET` commands on a fresh ubuntu 14.04 host machine.

### DIFF
--- a/common/scripts/x86_64/Ubuntu_14.04/remote/installRedis.sh
+++ b/common/scripts/x86_64/Ubuntu_14.04/remote/installRedis.sh
@@ -1,7 +1,11 @@
 #!/bin/bash -e
 
-REDIS_VERSION="4.0.8"
 TIMEOUT=60
+
+install_redis() {
+  echo "installing redis"
+  apt-get install -o Dpkg::Options::="--force-confold" --force-yes -y redis-server
+}
 
 set_defaults() {
   if [ -z "$REDIS_HOST" ]; then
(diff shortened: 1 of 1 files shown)


ric03uec:
please see how this is being done for u14 in current workflow and copy that file as is. i dont know why we're making changes to this script when we already have a script that works end to end

scriptnull:
Redis official documentation recommends installing from source. Hence, I wrote the script the way it is now.


Furthermore, if I try to use apt-get install like we do in u14all ( https://github.com/dry-dock/u14all/blob/master/version/redis_4_0.sh )

The redis server seem to be unreachable
```
root@975d9d6a0a2a:/admiral/common/scripts/x86_64/Ubuntu_14.04/remote# redis-cli 
Could not connect to Redis at 127.0.0.1:6379: Connection refused
Could not connect to Redis at 127.0.0.1:6379: Connection refused
```

ric03uec:
the rest of the script needs to change then. you cannot change the structure of the entire file. we need correct functions and naming conventions. if installation needs to change, only that part should change, nothing else

scriptnull:
Here is what we have in u14all images

```bash
#!/bin/bash -e

echo "================= Installing redis-server ==================="

sudo add-apt-repository ppa:chris-lea/redis-server
sudo apt-get update
sudo wget https://redis.io/download
sudo apt-get install -q -y redis-server=4:4.0.8*
```

Here is what we have in my script
```bash
#!/bin/bash -e

export REDIS_VERSION="4.0.8"

sudo apt-get update
sudo apt-get install -y build-essential tcl8.5
wget http://download.redis.io/releases/redis-$REDIS_VERSION.tar.gz
tar xzf redis-$REDIS_VERSION.tar.gz
pushd redis-$REDIS_VERSION
  make
  make test
  sudo make install
  sudo REDIS_PORT=6379 \
    REDIS_CONFIG_FILE=/etc/redis/6379.conf \
    REDIS_LOG_FILE=/var/log/redis_6379.log \
    REDIS_DATA_DIR=/var/lib/redis/6379 \
    REDIS_EXECUTABLE=`command -v redis-server` ./utils/install_server.sh
popd
```

Except `sudo apt-get update` nothing is common here.

> we need correct functions and naming conventions. 

We already don't have any functions wrapping the existing logic. Should I wrap my logic around a function called `install_redis_server()`?

> if installation needs to change, only that part should change, nothing else

Could you be more specific on what _part_ needs to change here?

ric03uec:
https://github.com/Shippable/admiral/blob/master/common/scripts/Ubuntu_14.04/installRedis.sh

ric03uec:
this is not a drydock update, its admiral script

scriptnull:
yeah ok. 

> please see how this is being done for u14 in current workflow and copy that file as is

I got confused that this statement meant to reuse the u14 script in drydock inside admiral. :sweat_smile: 

Is this the script that is going to call the OS specific `installRedis` script? 

https://github.com/Shippable/admiral/blob/f17d3ee5b112fd777f22be507980ea628e82b28f/common/scripts/installRedis.sh#L93-L95

If yes, could you tell me the way to test it? If not, we would need to set the value for `REDIS_HOST` and `REDIS_PORT` for our script to work.

https://github.com/Shippable/admiral/blob/f17d3ee5b112fd777f22be507980ea628e82b28f/common/scripts/Ubuntu_14.04/installRedis.sh#L38-L39

ric03uec:
no, this is the script that gets copied over to the remote host and gets executed there

Issue #1907, opened 2018-02-20
add swarm installMaster.sh for centOs


ChaitanyaVootla:
https://github.com/Shippable/pm/issues/10159

Issue #1906, opened 2018-02-20
add swarm installWorker.sh for centOs


ChaitanyaVootla:
https://github.com/Shippable/pm/issues/10160

PR #1908, opened 2018-02-20
adds swarm install files for centOs

ChaitanyaVootla:
fixes: https://github.com/Shippable/admiral/issues/1907
fixes: https://github.com/Shippable/admiral/issues/1906

### DIFF
--- a/common/scripts/x86_64/CentOS_7/installMaster.sh
+++ b/common/scripts/x86_64/CentOS_7/installMaster.sh
@@ -1,0 +1,40 @@
+#!/bin/bash -e
+
+export COMPONENT="master"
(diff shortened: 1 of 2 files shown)


common/scripts/x86_64/CentOS_7/installMaster.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash -e
```
Bharath92: why are we moving installMaster.sh to {arch}/{os}? this is not present in https://github.com/Shippable/admiral/tree/master/common/scripts/Ubuntu_14.04 and this file doesn't have any OS specific commands

Issue #1888, opened 2018-02-20
Move common/scripts/install_psql.sh to common/scripts/x86_64/Ubuntu_14.04/install_psql.sh. Use the updated path to install psql


Bharath92:
https://github.com/Shippable/pm/issues/10115

Bharath92:
this is done, closing

PR #1922, opened 2018-02-21
install psql in admiral image using x86_64/Ubuntu_14.04/install_psql.sh

Bharath92:
https://github.com/Shippable/admiral/issues/1888

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM drydock/microbase:master
 
 ADD . /home/shippable/admiral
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM drydock/microbase:master
 
 ADD . /home/shippable/admiral
 
-RUN ./home/shippable/admiral/common/scripts/install_psql.sh
+RUN ./home/shippable/admiral/common/scripts/x86_64/Ubuntu_14.04/install_psql.sh
```
ric03uec: can we have envs for os and arch

Issue #1947, opened 2018-02-22
Populate _installVersion.env for CentOS_7


shrivara:
https://github.com/Shippable/pm/issues/10194

PR #1952, opened 2018-02-22
Populate Docker, AWS CLI values for CentOS 7

shrivara:
https://github.com/Shippable/admiral/issues/1947


common/scripts/x86_64/CentOS_7/_installVersion.env
```diff
@@ -0,0 +1,2 @@
+readonly DOCKER_VERSION=17.06
+readonly AWSCLI_VERSION=1.11.91
```
shrivara: Same version as that of Ubuntu 16.04. This is the version [where `--no-include-email` support](https://github.com/aws/aws-cli/blob/develop/CHANGELOG.rst#11191) was added.

Issue #2043, opened 2018-02-27
Update the master installRedis.sh to copy the config files from new location.


jatindogra:
https://github.com/Shippable/pm/issues/10245

Issue #2042, opened 2018-02-27
Create a new file redis.conf in the config folder for OS U14 and U16.


jatindogra:
https://github.com/Shippable/pm/issues/10245

jatindogra:
Completed

PR #2044, opened 2018-02-27
adds redis.conf OS specific and updates installRedis.sh to call it from new path

jatindogra:
https://github.com/Shippable/admiral/issues/2042
https://github.com/Shippable/admiral/issues/2043


ric03uec:
lets remove all comments from config file

Issue #2047, opened 2018-02-28
Add and copy redis config for CentOS 7


shrivara:
https://github.com/Shippable/pm/issues/10239

PR #2048, opened 2018-02-28
Add redis conf for CentOS 7 and copy the config

shrivara:
https://github.com/Shippable/admiral/issues/2047

### DIFF
--- a/common/scripts/x86_64/CentOS_7/remote/config/redis.conf
+++ b/common/scripts/x86_64/CentOS_7/remote/config/redis.conf
@@ -1,0 +1,1053 @@
+# Redis configuration file example.
+#
+# Note that in order to read the configuration file, Redis must be
+# started with the file path as first argument:
+#
+# ./redis-server /path/to/redis.conf
+
+# Note on units: when memory size is needed, it is possible to specify
+# it in the usual form of 1k 5GB 4M and so forth:
+#
+# 1k => 1000 bytes
+# 1kb => 1024 bytes
+# 1m => 1000000 bytes
+# 1mb => 1024*1024 bytes
+# 1g => 1000000000 bytes
+# 1gb => 1024*1024*1024 bytes
+#
+# units are case insensitive so 1GB 1Gb 1gB are all the same.
+
+################################## INCLUDES ###################################
+
+# Include one or more other config files here.  This is useful if you
+# have a standard template that goes to all Redis servers but also need
+# to customize a few per-server settings.  Include files can include
+# other files, so use this wisely.
+#
+# Notice option "include" won't be rewritten by command "CONFIG REWRITE"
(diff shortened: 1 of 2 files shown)


common/scripts/x86_64/CentOS_7/remote/config/redis.conf
```diff
@@ -0,0 +1,1053 @@
...
+# instance to everybody on the internet. So by default we uncomment the
+# following bind directive, that will force Redis to listen only into
+# the IPv4 lookback interface address (this means Redis will be able to
+# accept connections only from clients running into the same computer it
+# is running).
+#
+# IF YOU ARE SURE YOU WANT YOUR INSTANCE TO LISTEN TO ALL THE INTERFACES
+# JUST COMMENT THE FOLLOWING LINE.
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# bind 127.0.0.1
```
shrivara: From the original file, this was commented out to allow external connections.

common/scripts/x86_64/CentOS_7/remote/config/redis.conf
```diff
@@ -0,0 +1,1053 @@
...
+#
+# The server only accepts connections from clients connecting from the
+# IPv4 and IPv6 loopback addresses 127.0.0.1 and ::1, and from Unix domain
+# sockets.
+#
+# By default protected mode is enabled. You should disable it only if
+# you are sure you want clients from other hosts to connect to Redis
+# even if no authentication is configured, nor a specific set of interfaces
+# are explicitly listed using the "bind" directive.
+protected-mode no
```
shrivara: This was changed to `no` to allow password-less connection.

common/scripts/x86_64/CentOS_7/remote/installRedis.sh
```diff
@@ -16,6 +16,9 @@ install_redis() {
   sudo yum install -y epel-release
   sudo yum update -y
   sudo yum install -y redis
+  # Move the config that we've copied earlier to the default
+  # location of the config in CentOS.
+  sudo mv /etc/redis/redis.conf /etc/redis.conf
```
jatindogra: Should we not also do systemctl daemon-reload?
I am not sure whether the changes to conf file will be reloaded or not.

shrivara: Hmm, didn't seem to need it why I did the install, but I'll have to confirm as I was experimenting.

shrivara: Actually I think we don't need to to, we're not changing the service definition, but the config that redis loads. So, we should be good.

Issue #2061, opened 2018-03-01
[bug] centos doesn't have /root/.ssh directory by default. this should be created


Bharath92:
https://github.com/Shippable/pm/issues/10310

Issue #2060, opened 2018-03-01
Generate ssh keys before parsing install args


Bharath92:
https://github.com/Shippable/pm/issues/10310

Bharath92:
this is done, closing

PR #2062, opened 2018-03-01
fixes adding public ssh key bugs

Bharath92:
1. generate ssh key before parsing install args
2. create /root/.ssh directory if not present(for centos)
3. do not delete public ssh key from authorized keys, if already present. adding same public ssh key multiple times doesn't throw error.

closes https://github.com/Shippable/admiral/issues/2061
https://github.com/Shippable/admiral/issues/2060

### DIFF
--- a/common/scripts/lib/_parseArgs.sh
+++ b/common/scripts/lib/_parseArgs.sh
@@ -312,17 +312,14 @@ __set_host_services() {
   echo "HOST_SERVICES=true" >> "$ADMIRAL_ENV"
   source "$ADMIRAL_ENV"
 
-  ## add ssh key to host's authorized keys
-  local authorized_keys="/root/.ssh/authorized_keys"
-  local tmp_authorized_keys="/root/.ssh/tmp"
+  # create .ssh dir, if not present
+  local root_ssh_dir="/root/.ssh"
+  mkdir -p "$root_ssh_dir"
+
+  ## add public ssh key to host's authorized keys
+  local authorized_keys="$root_ssh_dir/authorized_keys"
   local ssh_public_key=$(cat "$SSH_PUBLIC_KEY")
-  # remove public key from authorized_keys, if present
-  if grep -v "^$ssh_public_key" "$authorized_keys" > "$tmp_authorized_keys"; then
-    cat "$tmp_authorized_keys" > "$authorized_keys"
-  fi
-  rm -f "$tmp_authorized_keys"
-  # add public key to authorized_keys
-  echo "$ssh_public_key" >> "$authorized_keys"
+  echo "$ssh_public_key" | tee -a "$authorized_keys" &> /dev/null
 }
 
 __parse_args_install() {
@@ -544,8 +541,8 @@ __parse_args() {
     case $key in
       install)
         shift
-        __parse_args_install "$@"
         __generate_ssh_keys
+        __parse_args_install "$@"
         __accept_shippable_license
         ;;
       switch)
--- a/common/scripts/lib/_parseArgs.sh
+++ b/common/scripts/lib/_parseArgs.sh
@@ -308,18 +308,20 @@ __prompt_for_inputs() {
 }
 
 __set_host_services() {
-  sed -i '/^HOST_SERVICES/d' "$ADMIRAL_ENV"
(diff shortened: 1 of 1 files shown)


common/scripts/lib/_parseArgs.sh
```diff
@@ -312,17 +312,14 @@ __set_host_services() {
...
-  local authorized_keys="/root/.ssh/authorized_keys"
-  local tmp_authorized_keys="/root/.ssh/tmp"
+  # create .ssh dir, if not present
+  local root_ssh_dir="/root/.ssh"
+  mkdir -p "$root_ssh_dir"
+
+  ## add public ssh key to host's authorized keys
+  local authorized_keys="$root_ssh_dir/authorized_keys"
   local ssh_public_key=$(cat "$SSH_PUBLIC_KEY")
-  # remove public key from authorized_keys, if present
```
Bharath92: this is not required because adding same ssh key multiple times doesn't lead to any error. In centos, since the public ssh key is the only content of the `authorized_keys` file, inverse grep match (grep -v) returns error because grep doesn't find any other content to return. So, it exits with exit code 1. So instead of removing public key from authorized_keys, if present, we can just let public key to get added to authorized keys multiple times as it doesn't cause any error.

common/scripts/lib/_parseArgs.sh
```diff
@@ -308,21 +308,20 @@ __prompt_for_inputs() {
...
-  sed -i '/^HOST_SERVICES/d' "$ADMIRAL_ENV"
-  echo "HOST_SERVICES=true" >> "$ADMIRAL_ENV"
-  source "$ADMIRAL_ENV"
-
-  ## add ssh key to host's authorized keys
-  local authorized_keys="/root/.ssh/authorized_keys"
-  local tmp_authorized_keys="/root/.ssh/tmp"
-  local ssh_public_key=$(cat "$SSH_PUBLIC_KEY")
-  # remove public key from authorized_keys, if present
-  if grep -v "^$ssh_public_key" "$authorized_keys" > "$tmp_authorized_keys"; then
```
Bharath92: this causes a problem in centos because the centos file only has the public key admiral adds. grep inverse match (grep -v) will return an error because it doesn't find any other content than the ssh key added by admiral.

adding multiple ssh keys to authorized_keys doesn't cause any error. Even while adding public keys to authorized keys currently during multi-box install, we don't delete the public ssh key if already present. So, this logic of removing ssh key if present is removed.

common/scripts/lib/_parseArgs.sh
```diff
@@ -308,21 +308,20 @@ __prompt_for_inputs() {
...
+    source "$ADMIRAL_ENV"
+
+    # create .ssh dir, if not present
+    local root_ssh_dir="/root/.ssh"
+    mkdir -p "$root_ssh_dir"
+
+    ## add public ssh key to host's authorized keys
+    local authorized_keys="$root_ssh_dir/authorized_keys"
+    local ssh_public_key=$(cat "$SSH_PUBLIC_KEY")
+    echo "$ssh_public_key" | tee -a "$authorized_keys" &> /dev/null
```
Bharath92: We add ssh key [this way](https://github.com/Shippable/admiral/blob/master/common/scripts/x86_64/Ubuntu_14.04/_helpers.sh#L393), so following the pattern

common/scripts/lib/_parseArgs.sh
```diff
@@ -308,21 +308,20 @@ __prompt_for_inputs() {
...
-  if grep -v "^$ssh_public_key" "$authorized_keys" > "$tmp_authorized_keys"; then
-    cat "$tmp_authorized_keys" > "$authorized_keys"
+  if [ "$HOST_SERVICES" == "true" ]; then
+    sed -i '/^HOST_SERVICES/d' "$ADMIRAL_ENV"
+    echo "HOST_SERVICES=true" >> "$ADMIRAL_ENV"
+    source "$ADMIRAL_ENV"
+
+    # create .ssh dir, if not present
+    local root_ssh_dir="/root/.ssh"
+    mkdir -p "$root_ssh_dir"
```
Bharath92: this is required for centos

Issue #2056, opened 2018-02-28
Install db on host if HOST_SERVICES is set to true (common/scripts/installDb.sh)


Bharath92:
https://github.com/Shippable/pm/issues/10310

PR #2063, opened 2018-03-01
install db on host on centos, if HOST_SERVICES is true

Bharath92:
closes https://github.com/Shippable/admiral/issues/2056

### DIFF
--- a/common/scripts/x86_64/CentOS_7/remote/installDb.sh
+++ b/common/scripts/x86_64/CentOS_7/remote/installDb.sh
@@ -64,7 +64,7 @@ __update_sources() {
 
 __install_postgres() {
   echo "Installing Postgres"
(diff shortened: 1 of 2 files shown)


common/scripts/x86_64/CentOS_7/remote/installDb.sh
```diff
@@ -64,7 +64,7 @@ __update_sources() {
 
 __install_postgres() {
   echo "Installing Postgres"
-  sudo rpm -ivh http://yum.postgresql.org/9.5/redhat/rhel-7-x86_64/pgdg-centos95-9.5-2.noarch.rpm
+  sudo rpm -ivh --force http://yum.postgresql.org/9.5/redhat/rhel-7-x86_64/pgdg-centos95-9.5-3.noarch.rpm
```
Bharath92: we are installing 9.5.3 psql client. We should install similar postgres server

Issue #2068, opened 2018-03-06
DEV 10348.1  adds env in /etc/shippable/admiral.env to decide which version should be installed on master and workers


niranjan92:
https://github.com/Shippable/pm/issues/10348

niranjan92:
closing, this will be done in different issue

PR #2074, opened 2018-03-06
adds DOCKER_VERSION to admiral env template

niranjan92:
https://github.com/Shippable/admiral/issues/2068

adds env var `DOCKER_VER` to admiral.env as `DOCKER_VERSION` is already loaded from `_installVersion.sh`

if we put same variable, then it will get overwritten when we source `_installVersion.sh`


common/scripts/configs/admiral.env.template
```diff
@@ -31,3 +31,4 @@ SHIPPABLE_HTTPS_PROXY=""
 SHIPPABLE_NO_PROXY=""
 ARCHITECTURE=""
 OPERATING_SYSTEM=""
+SWARM_DOCKER_VERSION=""
```
ric03uec: just `DOCKER_VERSION` is good enough

niranjan92: we are already loading  `DOCKER_VERSION` environment variable from `_installVersion` for arch/OS
https://github.com/Shippable/admiral/blob/master/common/scripts/x86_64/Ubuntu_14.04/_installVersion.env#L1

should i use `INSTALLED_DOCKER_VERSION` instead of this?

Issue #2086, opened 2018-03-06
DEV 10348.1  for u14, upgrade should not fail if the docker version is not 1.13


niranjan92:
https://github.com/Shippable/pm/issues/10348

niranjan92:
this is done, closing

Issue #2071, opened 2018-03-06
DEV 10348.4 set an env INSTALLED_DOCKER_VERSION when admiral boots up


niranjan92:
https://github.com/Shippable/pm/issues/10348

niranjan92:
this is done, closing

PR #2092, opened 2018-03-06
docker upgrade should not fail if installed version is not 1.13

niranjan92:
https://github.com/Shippable/admiral/issues/2086
https://github.com/Shippable/admiral/issues/2071

- sets `INSTALLED_DOCKER_VERSION` env on boot and after installing docker
- tested admiral install and upgrade
- shows message if installed docker version is not same as admiral's docker version

### DIFF
--- a/common/scripts/lib/_helpers.sh
+++ b/common/scripts/lib/_helpers.sh
@@ -50,6 +50,14 @@ __check_os_and_architecture() {
   source "$ADMIRAL_ENV"
 }
 
+__set_installed_docker_version() {
+  if command -v "docker" &>/dev/null; then
+    docker_ver=$(docker version --format '{{.Server.Version}}')
+    export INSTALLED_DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')
+    echo "setting INSTALLED_DOCKER_VERSION=$INSTALLED_DOCKER_VERSION"
+  fi
+}
+
 __cleanup() {
   if [ -d $CONFIG_DIR ]; then
     __process_msg "Removing previously created $CONFIG_DIR"
--- a/common/scripts/x86_64/Ubuntu_14.04/_helpers.sh
+++ b/common/scripts/x86_64/Ubuntu_14.04/_helpers.sh
@@ -77,14 +77,10 @@ __check_dependencies() {
   ################## Install Docker  #####################################
   if type docker &> /dev/null && true; then
     __process_msg "'docker' already installed, checking version"
-    local docker_version=$(docker --version)
-    if [[ "$docker_version" == *"$DOCKER_VERSION"* ]]; then
-      __process_msg "'docker' $docker_version installed"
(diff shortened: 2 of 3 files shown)


common/scripts/x86_64/Ubuntu_14.04/_helpers.sh
```diff
@@ -77,17 +82,18 @@ __check_dependencies() {
...
+    local docker_version=$(docker version --format '{{.Server.Version}}')
+    if __version_gte "$docker_version" "$min_docker_version"; then
+      __process_msg "'docker' $docker_version version is greater than equal to $min_docker_version"
     else
-      __process_error "Docker version $docker_version installed, required $DOCKER_VERSION"
+      __process_error "Docker version $docker_version installed, min version required = $min_docker_version"
       __process_error "Install Docker Version $DOCKER_VERSION from \"
-      https://docs.docker.com/cs-engine/1.13/#install-on-ubuntu-1404-lts-or-1604-lts"
+      https://docs.docker.com/v17.06/engine/installation/linux/ "
       exit 1
```
ric03uec: - will this not exit?
- also, according to the design, we need to set the docker version in the ENV and then use that env for any comparisions. this implementation is incorrect
- secondly, why are we complicating this? if the installed version is not equal to desired version, just print a message and proceed. not sure why we need less than/greater than check here

common/scripts/lib/_helpers.sh
```diff
@@ -50,6 +50,14 @@ __check_os_and_architecture() {
   source "$ADMIRAL_ENV"
 }
 
+__set_installed_docker_version() {
+  if command -v "docker" &>/dev/null; then
```
ric03uec: will thsi not fail if docker is not installed?

common/scripts/x86_64/Ubuntu_14.04/_helpers.sh
```diff
@@ -77,14 +77,11 @@ __check_dependencies() {
...
-      __process_msg "'docker' $docker_version installed"
+    if [[ "$INSTALLED_DOCKER_VERSION" == *"$DOCKER_VERSION"* ]]; then
+      __process_msg "'docker' $INSTALLED_DOCKER_VERSION installed"
     else
-      __process_error "Docker version $docker_version installed, required $DOCKER_VERSION"
-      __process_error "Install Docker Version $DOCKER_VERSION from \"
-      https://docs.docker.com/cs-engine/1.13/#install-on-ubuntu-1404-lts-or-1604-lts"
-      exit 1
+      __process_msg "Installed Docker version - $INSTALLED_DOCKER_VERSION \
+      not same as expected Docker version - $DOCKER_VERSION."
```
ric03uec: `... not the same as latest supported version ...`

common/scripts/lib/_helpers.sh
niranjan92: this checks if `docker` is command is present before calling `docker version` [reference](https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script)

Issue #2085, opened 2018-03-06
 Set defaultTaskImage for Windows runtimeTemplates to w16


scriptnull:
https://github.com/Shippable/pm/issues/10308

scriptnull:
Verified on RC.

PR #2093, opened 2018-03-06
uses w16 instead of w16microbase

scriptnull:
https://github.com/Shippable/admiral/issues/2085


common/scripts/configs/create_master_runtime_templates.sql
```diff
@@ -20,9 +20,10 @@ do $$
       end if;
 
       -- x86_64 WindowsServer_2016 Master
+      update "runtimeTemplates" set "defaultTaskImage" = 'w16' where "archTypeCode" = 8000 and "osTypeCode" = 9002 and "version" = 'Master';
```
ric03uec: needs an if condition

scriptnull: I guess `where` is an equivalent of `if` here. `where` will only update the rows that satisfies the condition.
```
where "archTypeCode" = 8000 and "osTypeCode" = 9002 and "version" = 'Master';
```

ric03uec: what if runtime templates is not there for old installations?

scriptnull: Execution enters this block only if `runtimeTemplates` table is present in the installation ( https://github.com/Shippable/admiral/blob/master/common/scripts/configs/create_master_runtime_templates.sql#L3 )

So, there shouldn't be any problem if the `runtimeTemplates` table does not exist or row for windows is not present in `runtimeTemplates`

Issue #2144, opened 2018-03-11
Pull Shippable service images on swarm master and workers on upgrade.


Bharath92:
https://github.com/Shippable/pm/issues/10352

Bharath92:
this is done, closing

Issue #2143, opened 2018-03-11
Add `pull_image_master` function to pull Shippable service images on master machine


Bharath92:
https://github.com/Shippable/pm/issues/10352

Bharath92:
this is done, closing

Issue #2142, opened 2018-03-11
Pull Shippable service images on the swarm master machine during install.


Bharath92:
https://github.com/Shippable/pm/issues/10352

Bharath92:
this is done, closing

Issue #2141, opened 2018-03-11
Pass `PRIVATE_IMAGE_REGISTRY` and `RELEASE` environment variables to the `installMaster.sh` script


Bharath92:
https://github.com/Shippable/pm/issues/10352

Bharath92:
this is done, closing

PR #2146, opened 2018-03-12
pulls service images only when required

Bharath92:
- install pulls admiral image always
- install --dev pulls stateful service images always
- init master/worker pulls Shippable service images
- upgrade pulls only Shippable service images on master & worker

https://github.com/Shippable/admiral/issues/2139
https://github.com/Shippable/admiral/issues/2140
https://github.com/Shippable/admiral/issues/2141
https://github.com/Shippable/admiral/issues/2142
https://github.com/Shippable/admiral/issues/2143
https://github.com/Shippable/admiral/issues/2144

verified by doing
- a fresh onebox install
- fresh dev install
- upgrade onebox install

### DIFF
--- a/common/scripts/lib/_helpers.sh
+++ b/common/scripts/lib/_helpers.sh
@@ -7,7 +7,8 @@
 # files after we add release and re-install features
 
 declare -a SERVICE_IMAGES=("api" "www" "micro" "mktg" "nexec" "genexec")
-declare -a PRIVATE_REGISTRY_IMAGES=("admiral" "postgres" "vault" "rabbitmq" "gitlab" "redis")
+declare -a PRIVATE_REGISTRY_IMAGES=("postgres" "vault" "rabbitmq" "gitlab" "redis")
+export ADMIRAL_IMAGE=admiral""
 
 __bootstrap_admiral_env() {
   ################## load env file  ###########################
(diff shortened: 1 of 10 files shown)


common/scripts/lib/_helpers.sh
```diff
@@ -7,7 +7,8 @@
 # files after we add release and re-install features
 
 declare -a SERVICE_IMAGES=("api" "www" "micro" "mktg" "nexec" "genexec")
-declare -a PRIVATE_REGISTRY_IMAGES=("admiral" "postgres" "vault" "rabbitmq" "gitlab" "redis")
+declare -a PRIVATE_REGISTRY_IMAGES=("postgres" "vault" "rabbitmq" "gitlab" "redis")
+export ADMIRAL_IMAGE=admiral""
```
ric03uec: double quotes are not surrounding the text

common/scripts/x86_64/CentOS_7/_helpers.sh
```diff
@@ -339,6 +343,57 @@ __pull_images() {
...
+  fi
+
+  for image in "${SERVICE_IMAGES[@]}"; do
+    image="$PRIVATE_IMAGE_REGISTRY/$image:$RELEASE"
+    local pull_cmd="sudo docker pull $image"
+    if [ -z "$master_ip" ]; then
+      __process_msg "Pulling $image"
+      sudo docker pull $image
+    else
+      local docker_login_cmd="aws ecr --no-include-email --region us-east-1 get-login | bash"
```
ric03uec: why are we not using the docker login helper function to do this. again we'll end up in a situation where this logic is scattered all across the codebase and cannot be changed.

ric03uec: my bad. this executes on remote thats why i suppose. But again, the command that gets constructed here should be centralized in one function since we use it in multiple places

Bharath92: this taken from the previous flow of how we were doing it. I would like to clean it, but because of time contraint I implemented whatever was there before. Probably, I could do it after this release?

ric03uec: ok. lets make a note of it then here under a `TODO`. we should try to clean things up with subsequent iterations on this file.

Bharath92: added

PR #2152, opened 2018-03-12
initialize swarm master machine

Bharath92:
https://github.com/Shippable/admiral/issues/2144
https://github.com/Shippable/admiral/issues/2142

since we use ssh to pull images on the master, `rc_deploy` was failing because `/root/.aws/credentials` file was not present. We just add the aws credentials file in ~/.aws. Now it is required that it should be present in `/root/.aws/credentials` since we use `root@localhost` to ssh into the machine.

We will be adding the credentials to `/root/.aws` as part of initializing the swarm master.
For backward compatibility, for users who have currently installed Shippable, and are willing to upgrade, we need to explicitly add the credentials to `/root/.aws` since they have already initialized their swarm masters.


common/scripts/x86_64/Ubuntu_14.04/_helpers.sh
```diff
@@ -227,7 +227,9 @@ __registry_login() {
     sed -i "s#{{SECRET_KEY}}#$SECRET_KEY#g" $credentials_file
 
     mkdir -p ~/.aws
-    mv -v $credentials_file ~/.aws
+    mkdir -p /root/.aws
```
Bharath92: for users who want to upgrade

Issue #2151, opened 2018-03-12
10414 DEV Create folder structures and add scripts to bring up the Admiral container on a fresh RHEL 7 machine 


Riturajcse:
https://github.com/Shippable/pm/issues/10414

Riturajcse:
This is complete.
Closing

PR #2154, opened 2018-03-12
adds missing functions for RHEL_7

Riturajcse:
https://github.com/Shippable/admiral/issues/2151

### DIFF
--- a/common/scripts/x86_64/RHEL_7/_helpers.sh
+++ b/common/scripts/x86_64/RHEL_7/_helpers.sh
@@ -1,5 +1,9 @@
 #!/bin/bash -e
 
+declare -a SERVICE_IMAGES=("api" "www" "micro" "mktg" "nexec" "genexec")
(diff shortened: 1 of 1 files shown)


common/scripts/x86_64/RHEL_7/_helpers.sh
```diff
@@ -340,6 +344,58 @@ __pull_images() {
...
+
+  for image in "${SERVICE_IMAGES[@]}"; do
+    image="$PRIVATE_IMAGE_REGISTRY/$image:$RELEASE"
+    local pull_cmd="sudo docker pull $image"
+    if [ -z "$master_ip" ]; then
+      __process_msg "Pulling $image"
+      sudo docker pull $image
+    else
+      # TODO: have a function which returns the docker login command instead of figuring it out everytime
+      local docker_login_cmd="aws ecr --no-include-email --region us-east-1 get-login | bash"
```
shrivara: This needs to also support the `--no-ssl-verify` flag. Something similar to https://github.com/Shippable/admiral/blob/537c1549802c9876e4d3906f803213e37cca5672/common/scripts/lib/_helpers.sh#L246.

Issue #2156, opened 2018-03-12
bring up swarm services on host network


Bharath92:
https://github.com/Shippable/pm/issues/10352

Bharath92:
this is done, closing

PR #2157, opened 2018-03-12
bring up services on host network

Bharath92:
https://github.com/Shippable/admiral/issues/2156
this brings up services on `--network host`. So, we don't need the private IP to connect to the machines. We can use localhost to talk to the services.

also, fixed a couple of bugs
- if docker version is 1.13, the ingress network should be used
- if NO_VERIFY_SSL flag is present appropriate docker login command should be used

### DIFF
--- a/api/services/apiConfig.js
+++ b/api/services/apiConfig.js
@@ -249,7 +249,7 @@ function _generateRunCommandCluster(bag, next) {
 
   var opts;
 
-  if (bag.installedDockerVersion === '1.13')
+  if (bag.installedDockerVersion === 1.13)
     opts = util.format(' --publish mode=host,target=%s,published=%s' +
       ',protocol=tcp' +
       ' --network ingress' +
@@ -259,6 +259,7 @@ function _generateRunCommandCluster(bag, next) {
   else
     opts = util.format(' --publish mode=host,target=%s,published=%s' +
       ',protocol=tcp' +
+      ' --network host' +
       ' --mode global' +
       ' --with-registry-auth' +
       ' --endpoint-mode vip', bag.port, bag.port);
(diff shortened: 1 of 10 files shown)


api/services/apiConfig.js
```diff
@@ -249,7 +249,7 @@ function _generateRunCommandCluster(bag, next) {
 
   var opts;
 
-  if (bag.installedDockerVersion === '1.13')
+  if (bag.installedDockerVersion === 1.13)
```
ric03uec: we should keep this as string comparison. versions should not be treated as floating point decimals. if the === isnt working, use underscore methods for comparison which will sure that this doesnt change even if the version becomes something else

Bharath92: we get the version as a number here. we do JSON.parse('1.13'), which returns 1.13 as a number. So, i think we should do number comparision

ric03uec: this just means the logic will change if we replace 1.13 with 17.12.1-ce (or something that is strictly not a number). the solution should be generic enough (within sane bounds) to facilitate foreseeable use cases. this one does not do it. other points:
- the version should be stored up top in `bag.deprecatedDockerVersion`
- once set, this variable should be used throughout the script
- the comparison should be agnostic of version type so if json.parse is returning integer, convert it to string(or use underscore that'll take care of this mess)

Bharath92: I think it is already happening. The value of docker version is only up to 2 decimal places currently. If the instlalled docker version is 1.13 then 1.13 is stored. If the docker version is 17.06.0-ce, then 17.06 will be stored.

the only thing wrong here is the check, that is what is fixed

ric03uec: - version is not getting set in `bag` right now
- to reiterate what i said, versions cannot be treated as integers.  `json.parse('12.10')` will return `12.1` which is incorrect. the problem with this code is not the comparison but the underlying assumption that version is (and always will be) a real number

Issue #2195, opened 2018-03-15
10454 DEV Update admiral upgrade flow to upgrade system node components


Riturajcse:
https://github.com/Shippable/pm/issues/10454

Riturajcse:
This is complete.
Closing.

PR #2204, opened 2018-03-16
manages local node component during upgrade

Riturajcse:
https://github.com/Shippable/admiral/issues/2195

- Added separate files for the functions which were OS specific.
- Verified by first running admiral and then initializing all the services  from UI,
  - running `./manageSharedNode add` was able to start all the components
  - running `./manageSharedNode clean` was able to remove all the components
  - upgrade will first check for `DEV_MODE` and then will clean shared nodes and then add them again


common/scripts/upgrade.sh
```diff
@@ -410,6 +410,30 @@ __cleanup_workers() {
   fi
 }
 
+export system_cluster=null
```
ric03uec: all exports should be on top of the file

common/scripts/upgrade.sh
```diff
@@ -410,6 +411,29 @@ __cleanup_workers() {
...
+  __process_marker "getting systemClusters"
+  _shippable_get_systemClusters 'isDefault=true'
+  echo $response
+  system_cluster=$(echo $response \
+    | jq -r '.[0]')
+  if [ "$(echo $response | jq -r '.[0]')" != null ]; then
+    is_default_system_cluster_available=true
+  fi
+
+   __process_msg "is_default_system_cluster_available: $is_default_system_cluster_available"
```
Bharath92: 3 spaces

common/scripts/lib/_parseArgs.sh
```diff
@@ -600,6 +600,7 @@ __wipe_clean() {
       for id in ${container_ids[@]}; do
         eval "$remove_cmd $id"
       done
+      sudo rm -rf /var/lib/shippable/shippable_dev
```
shrivara: Shouldn't this do the full `./manageSharedNode clean` flow?

common/scripts/upgrade.sh
```diff
@@ -410,6 +411,29 @@ __cleanup_workers() {
...
+}
+
+__get_systemClusters() {
+  __process_marker "getting systemClusters"
+  _shippable_get_systemClusters 'isDefault=true'
+  echo $response
+  system_cluster=$(echo $response \
+    | jq -r '.[0]')
+  if [ "$(echo $response | jq -r '.[0]')" != null ]; then
+    is_default_system_cluster_available=true
```
shrivara: Is this only used for output?

common/scripts/upgrade.sh
```diff
@@ -410,6 +411,29 @@ __cleanup_workers() {
...
+__upgrade_local_system_node_components() {
+  __process_marker "upgrading local system nodes"
+  __get_systemClusters
+  if [ "$(echo $system_cluster | jq -r '.')" != null ]; then
+    sudo ./manageSharedNode clean
+    sudo ./manageSharedNode add
+  fi
+}
+
+__get_systemClusters() {
```
shrivara: We should probably name this something like `get_default_systemCluster`. The current name suggests that we are getting all `systemClusters`.

manageSharedNode
```diff
@@ -266,6 +266,7 @@ __setup_dirs() {
 
 __setup_mounts() {
   __process_marker "setting mounts"
+  source "$LOCAL_SCRIPTS_DIR/$ARCHITECTURE/$OPERATING_SYSTEM/setup_mounts.sh"
   /bin/bash -c "$LOCAL_SCRIPTS_DIR/$ARCHITECTURE/$OPERATING_SYSTEM/setup_mounts.sh"
```
shrivara: This isn't necessary anymore.

Issue #2211, opened 2018-03-16
Add logic to not mount docker binary from host to admiral container


Bharath92:
https://github.com/Shippable/pm/issues/10352

Bharath92:
this is done, closing

Issue #2210, opened 2018-03-16
Add the static docker binary while building the admiral image


Bharath92:
https://github.com/Shippable/pm/issues/10352

Bharath92:
this is done, closing

PR #2212, opened 2018-03-16
package and use the static binary in admiral image

Bharath92:
https://github.com/Shippable/admiral/issues/2211
https://github.com/Shippable/admiral/issues/2210

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN cd /home/shippable/admiral && npm install --unsafe-perm && \
 
 RUN mkdir -p /var/run/shippable/logs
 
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -9,5 +9,10 @@ RUN cd /home/shippable/admiral && npm install --unsafe-perm && \
 
 RUN mkdir -p /var/run/shippable/logs
 
+RUN apt-get update && wget https://download.docker.com/linux/static/stable/x86_64/docker-17.06.0-ce.tgz -O /tmp/docker.tar.gz && \
```
shrivara: I think the apt-get update is unnecessary.

Issue #2216, opened 2018-03-19
Run installWorker.sh on the workers on upgrade for existing customers


Bharath92:
https://github.com/Shippable/pm/issues/10352

Bharath92:
this is done, closing

Issue #2190, opened 2018-03-15
[bug] admiral.sh restart doesn't work on onebox install


Bharath92:
https://github.com/Shippable/pm/issues/10352

This is happening because we have a check which tries to start the db container, if DB_IP and ADMIRAL_IP are same. If the container is not found, the restart script exits with the error that db container is not found. Since, db is now installed on host, we should continue with the flow if db container is not found if DB_IP and ADMIRAL_IP are same, like we do for multi-box install

Bharath92:
this is done, closing

PR #2217, opened 2018-03-19
runs installWorker.sh script while upgrading docker 1.13 to docker 17.06 for existing customers

Bharath92:
https://github.com/Shippable/admiral/issues/2211
https://github.com/Shippable/admiral/issues/2190
https://github.com/Shippable/admiral/issues/2216

### DIFF
--- a/common/scripts/restart.sh
+++ b/common/scripts/restart.sh
@@ -16,6 +16,7 @@ __check_admiral() {
 }
 
 __start_secrets() {
+  local secrets_container_name="secrets"
   __process_marker "Starting secrets"
   __process_msg "Getting secrets config from DB"
   local secrets=""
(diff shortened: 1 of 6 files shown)


common/scripts/restart.sh
```diff
@@ -33,10 +33,10 @@ __start_secrets() {
 
   if [ "$secrets_ip" == "$ADMIRAL_IP" ]; then
     __process_msg "Checking secrets container"
-    local secrets_container=$(sudo docker ps -a -q -f "name=secrets" | awk '{print $1}')
-    if [ "$secrets_container" != "" ]; then
+    local secrets_container=$(sudo docker ps -a --format "{{.Names}}" | grep -w "secrets") || true
+    if [ ! -z "$secrets_container" ]; then
       __process_msg "Found a stopped secrets container, starting it"
-      sudo docker start $secrets_container
+      sudo docker start secrets
```
ric03uec: can we make `secrets` as a variable then use it in the grep and docker start command. ideally the secrets image name, container name should be stored up top but if that requires lot of changes, the least we should do it make the functions clean

Issue #2232, opened 2018-03-20
DEV update migrations.sql to add column maxDiskUsagePercentage


niranjan92:
https://github.com/Shippable/pm/issues/10458

niranjan92:
verified on RC, closing

PR #2234, opened 2018-03-20
adds max disk usage to systemSettings

niranjan92:
https://github.com/Shippable/admiral/issues/2232

syssettings


migrations/migrations.sql
```diff
@@ -5053,6 +5053,10 @@ do $$
       alter table "accounts" drop column "lastHubspotSyncAt";
     end if;
 
+    if not exists(select 1 from information_schema.columns where table_name = 'systemSettings' and column_name = 'maxDiskUsagePercentage') then
```
shrivara: We add `systemSettings` changes in [here](https://github.com/Shippable/admiral/blob/master/common/scripts/configs/system_settings.sql.template).

Issue #2245, opened 2018-03-22
10475 DEV Start nf.jira service if the Jira account integration is enabled in Admiral 


Riturajcse:
https://github.com/Shippable/pm/issues/10475

Riturajcse:
This is complete.
Closing

PR #2255, opened 2018-03-23
adds routeRoles for jira issue post passthrough

Riturajcse:
https://github.com/Shippable/admiral/issues/2245

- Adds `propertyBag` in runs schema
- Adds routeRoles for new jira issue create passthrough route
- Adds `nf.jira` queue 
- Eables jira microService when jira integration in enabled.

### DIFF
--- a/migrations/migrations.sql
+++ b/migrations/migrations.sql
@@ -391,6 +391,11 @@ do $$
     if not exists (select 1 from information_schema.columns where table_name = 'runs' and column_name = 'isPipelineTriggered') then
       alter table "runs" add column "isPipelineTriggered" BOOLEAN;
     end if;
+   
+    -- Adds runs propertyBag
+    if not exists (select 1 from information_schema.columns where table_name = 'runs' and column_name = 'propertyBag') then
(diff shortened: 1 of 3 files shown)


migrations/migrations.sql
```diff
@@ -391,6 +391,11 @@ do $$
     if not exists (select 1 from information_schema.columns where table_name = 'runs' and column_name = 'isPipelineTriggered') then
       alter table "runs" add column "isPipelineTriggered" BOOLEAN;
     end if;
+   
+    -- Adds runs propertyBag
+    if not exists (select 1 from information_schema.columns where table_name = 'runs' and column_name = 'propertyBag') then
+      alter table "runs" add column "propertyBag" TEXT;
```
ric03uec: this will lock up the whole table. we cannot run this query directly. a better way would be to create a new table with the new column, add the data from old to new table and rename new table.  
@rageshkrishna im not merging this right now. pl take a look when you get a chance

Riturajcse: Just FYI we are not populating any default value so I don't think this should take much time.

Issue #2268, opened 2018-03-28
10519 DEV update admiral/api/services/wwwConfig.js to set SHIPPABLE_PUBLIC_API_URL and SHIPPABLE_INT_API_URL


ankul-shippable:
https://github.com/Shippable/pm/issues/10519

ankul-shippable:
this is rc verified, closing.

PR #2269, opened 2018-03-28
update admiral/api/services/wwwConfig.js to set SHIPPABLE_PUBLIC_API_URL and SHIPPABLE_INT_API_URL

ankul-shippable:
#2268

```
 "Env": [
                "SHIPPABLE_API_URL=http://172.17.0.1:50000",
                "SHIPPABLE_PUBLIC_API_URL=http://172.17.0.1:50000",
                "SHIPPABLE_INT_API_URL=http://172.17.0.1:50000",
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "BASH_ENV=/etc/drydock/.env"
            ],

```

### DIFF
--- a/api/services/wwwConfig.js
+++ b/api/services/wwwConfig.js
@@ -123,12 +123,10 @@ function _getAPISystemIntegration(bag, next) {
             'No systemIntegration found for query: ' + query)
         );
 
-      var defaultAPIIntegration = _.findWhere(
+      bag.publicApiIntegration = _.findWhere(
         systemIntegrations, {name: 'api'});
-      var serviceAPIIntegration = _.findWhere(
-        systemIntegrations, {name: bag.config.apiUrlIntegration});
-
-      bag.apiSystemIntegration = serviceAPIIntegration || defaultAPIIntegration;
+      bag.internalApiIntegration = _.findWhere(
+        systemIntegrations, {name: 'internalAPI'});
 
       return next();
     }
@@ -149,10 +147,15 @@ function _generateEnvs(bag, next) {
   var who = bag.who + '|' + _generateEnvs.name;
   logger.verbose(who, 'Inside');
 
-  var apiUrl = bag.apiSystemIntegration.data &&
-    bag.apiSystemIntegration.data.url;
+  var publicApiUrl = bag.publicApiIntegration.data &&
+    bag.publicApiIntegration.data.url;
 
-  if (!apiUrl)
+  var internalApiUrl = bag.internalApiIntegration.data &&
+    bag.internalApiIntegration.data.url;
+
+  internalApiUrl = internalApiUrl || publicApiUrl;
+
+  if (!publicApiUrl)
     return next(
       new ActErr(who, ActErr.OperationFailed,
         'No apiUrl found.')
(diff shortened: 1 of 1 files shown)


api/services/wwwConfig.js
```diff
@@ -162,7 +165,11 @@ function _generateEnvs(bag, next) {
   envs = util.format('%s -e %s=%s',
     envs, 'SHIPPABLE_API_TOKEN', bag.serviceUserToken);
   envs = util.format('%s -e %s=%s',
-    envs, 'SHIPPABLE_API_URL', apiUrl);
+    envs, 'SHIPPABLE_API_URL', publicApiUrl);
```
ankul-shippable: will remove this once www is updated to use new ENVs

api/services/wwwConfig.js
```diff
@@ -123,12 +123,10 @@ function _getAPISystemIntegration(bag, next) {
             'No systemIntegration found for query: ' + query)
         );
 
-      var defaultAPIIntegration = _.findWhere(
+      bag.publicApiIntegration = _.findWhere(
         systemIntegrations, {name: 'api'});
-      var serviceAPIIntegration = _.findWhere(
```
shrivara: Do we want to remove this? I'm not quite sure what the original purpose of this was.

api/services/wwwConfig.js
```diff
@@ -149,10 +147,15 @@ function _generateEnvs(bag, next) {
   var who = bag.who + '|' + _generateEnvs.name;
   logger.verbose(who, 'Inside');
 
-  var apiUrl = bag.apiSystemIntegration.data &&
-    bag.apiSystemIntegration.data.url;
+  var publicApiUrl = bag.publicApiIntegration.data &&
+    bag.publicApiIntegration.data.url;
+
+  var internalApiUrl = bag.internalApiIntegration.data &&
```
shrivara: Is `bag.internalApiIntegration` always defined?

api/services/wwwConfig.js
```diff
@@ -123,12 +123,10 @@ function _getAPISystemIntegration(bag, next) {
             'No systemIntegration found for query: ' + query)
         );
 
-      var defaultAPIIntegration = _.findWhere(
+      bag.publicApiIntegration = _.findWhere(
         systemIntegrations, {name: 'api'});
-      var serviceAPIIntegration = _.findWhere(
-        systemIntegrations, {name: bag.config.apiUrlIntegration});
-
-      bag.apiSystemIntegration = serviceAPIIntegration || defaultAPIIntegration;
```
Bharath92: i guess u can keep this. `bag.internalApiIntegration = bag.internalApiIntegration || bag.publicApiIntegration;`

api/services/wwwConfig.js
Bharath92: why remove this? `SHIPPABLE_API_URL` is public api url right? Is the reason, to make public api url env more explicit?

api/services/wwwConfig.js
ankul-shippable: I think this was there to make sure www uses the api Url from the systemIntegration defined in config.json

api/services/wwwConfig.js
ankul-shippable: ok, I'll update

api/services/wwwConfig.js
ankul-shippable: yeah, it will always be defined

api/services/wwwConfig.js
ankul-shippable: yes, we want to make it more explicit to the users.

api/services/wwwConfig.js
```diff
@@ -149,10 +149,15 @@ function _generateEnvs(bag, next) {
...
 
-  var apiUrl = bag.apiSystemIntegration.data &&
-    bag.apiSystemIntegration.data.url;
+  var publicApiUrl = bag.publicApiIntegration.data &&
+    bag.publicApiIntegration.data.url;
+
+  var internalApiUrl = bag.internalApiIntegration.data &&
+    bag.internalApiIntegration.data.url;
+
+  internalApiUrl = internalApiUrl || publicApiUrl;
```
shrivara: This is probably not necessary any more.

ankul-shippable: updated

Issue #2280, opened 2018-03-29
10524 DEV Update api/systemMachineImages/post.js and put.js to handle runtimeTemplateId


deepikasl:
https://github.com/Shippable/pm/issues/10524

deepikasl:
This is done.

Issue #2277, opened 2018-03-29
10524 DEV Update scripts/dashboard/dashboardNew to support adding/editing runtimeTemplateId from Add New SMI and Edit SMI form


deepikasl:
https://github.com/Shippable/pm/issues/10524

deepikasl:
This is done.

PR #2281, opened 2018-03-29
add runtime version field in add/edit smi

deepikasl:
https://github.com/Shippable/admiral/issues/2280,
https://github.com/Shippable/admiral/issues/2277

Adding SMI:
anim

### DIFF
--- a/static/scripts/dashboard/dashboardNewCtrl.js
+++ b/static/scripts/dashboard/dashboardNewCtrl.js
@@ -23,7 +23,9 @@
       gitlabKeys: 'gitlab'
     };
     var systemMachineImagesById;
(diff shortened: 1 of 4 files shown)


static/scripts/dashboard/dashboardNewCtrl.js
```diff
@@ -2642,7 +2663,7 @@
     }
 
     function removeWorker (worker) {
-      if (worker.isInitialized)
+      if (worker.isInitialized || worker.isFailed)
```
deepikasl: this is for: deleting a swarm worker which is failed for the first time (swarm worker is not initialized) should be added to deletedWorkers.

Issue #2328, opened 2018-04-16
10678 DEV remove moveToGrisham route references from admiral


ankul-shippable:
https://github.com/Shippable/pm/issues/10678

ankul-shippable:
this is done, closing.

PR #2329, opened 2018-04-16
remove moveToGrisham route references from admiral

ankul-shippable:
#2328


ric03uec:
looks good. will be merged after the deployment tonight

PR #2410, opened 2018-05-08
adds apt-get update before installing packages

Bharath92:
https://github.com/Shippable/admiral/issues/2409


common/scripts/x86_64/Ubuntu_16.04/_helpers.sh
```diff
@@ -13,6 +13,8 @@ export ADMIRAL_IMAGE="u16admiral"
 __check_dependencies() {
   __process_marker "Checking dependencies"
 
+  apt-get update
```
ric03uec: lets add a `-y` just to be safe

Bharath92: i dont think this is required for update

Issue #2416, opened 2018-05-08
Show a SAAS-only panel to allow addition of Google Cloud Key in addition to AWS keys when configuring on-demand node pools


shrivara:
https://github.com/Shippable/pm/issues/10779

PR #2417, opened 2018-05-08
Support on-demand node pool gcloudKey provider

shrivara:
on-demand-node-pools-a
on-demand-node-pools-b
smi

https://github.com/Shippable/admiral/issues/2416

### DIFF
--- a/static/scripts/dashboard/dashboardNew.html
+++ b/static/scripts/dashboard/dashboardNew.html
@@ -2175,174 +2175,206 @@
                 </div>
                 <div id="on-demand-nodes" class="panel-collapse {{vm.initialized ? 'collapse' : 'collapse in'}}">
                   <div class="portlet-body">
+
+                    <!--
+                      Keys for AWS dynamic nodes
+                    -->
                     <div class="row">
                       <div class="col-sm-12">
                         <div class="card-box">
+                          <h4 class="m-t-0 header-title"><b>AWS</b></h4>
+                          <br />
                           <div class="toggle-bar">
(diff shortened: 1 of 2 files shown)


static/scripts/dashboard/dashboardNew.html
```diff
@@ -2175,174 +2175,206 @@ <h3 class="portlet-title text-dark">
...
-                            </div>
+                        </div>
+                      </div>
+                    </div>
+
+                    <!--
+                      Keys for GCloud dynamic nodes
+                      TODO: Enable this for server installs once things calm down.
+                    -->
+                    <div class="row" ng-if="vm.admiralEnv.IS_SERVER === false">
```
shrivara: Shown only in SAAS for now.

static/scripts/dashboard/dashboardNewCtrl.js
```diff
@@ -3677,14 +3684,27 @@
...
+            data: provisionIntegration.data,
+            isEnabled: provisionIntegration.isEnabled
+          };
 
-      updateSystemIntegration(bag,
+          // TODO: Remove this once all the references in the system
+          // have been updated to correctly handle multiple "provision"
+          // providers.
+          if (bag.masterName === 'gcloudKey')
+            bag.name = 'provision-gcloud';
```
shrivara: We have a few places in API which refer to provision integration without the `masterName` which will get confused when we have two such integrations, these references will get cleaned up along the way. Once we have cleaned it all up, this will be back to provision.

Issue #2486, opened 2018-06-21
10993 DEV Add migrations to remove unique constraints from the resource name subscriptionId index


Riturajcse:
https://github.com/Shippable/pm/issues/10993

Riturajcse:
![image](https://user-images.githubusercontent.com/14016521/41704937-369d0630-7555-11e8-8dd1-772194e32383.png)

This is complete.
Closing.

Riturajcse:
![image](https://user-images.githubusercontent.com/14016521/42150700-355734c0-7df8-11e8-881c-e31b1277801f.png)

This is complete.
Closing

Riturajcse:
This is complete.
Closing

PR #2488, opened 2018-06-21
adds migrations to reindex resourceSubscriptionIdNameU to resourceSubscriptionIdNameBranchU in resources table

Riturajcse:
https://github.com/Shippable/admiral/issues/2486

image

### DIFF
--- a/migrations/migrations.sql
+++ b/migrations/migrations.sql
@@ -809,6 +809,15 @@ do $$
       alter table "resources" add column "branch" varchar(255);
     end if;
 
+    -- Reindex resourceSubscriptionIdNameU to resourceSubscriptionIdNameBranchU in resources table
+    if exists (select 1 from pg_indexes where tablename = 'resources' and indexname = 'resourceSubscriptionIdNameU') then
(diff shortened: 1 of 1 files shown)


migrations/migrations.sql
```diff
@@ -809,6 +809,15 @@ do $$
       alter table "resources" add column "branch" varchar(255);
     end if;
 
+    -- Reindex resourceSubscriptionIdNameU to resourceSubscriptionIdNameBranchU in resources table
+    if exists (select 1 from pg_indexes where tablename = 'resources' and indexname = 'resourceSubscriptionIdNameU') then
+      drop index "resourceSubscriptionIdNameU";
+    end if;
+
+    if not exists (select 1 from pg_indexes where tablename = 'resources' and indexname = 'resourceSubscriptionIdNameBranchU') then
```
shrivara: I think lets switch the order of the migrations. At least one of these constraints will exist at all times then. Also, I think we should add this in the launch tasks so the column/index are created before the deployment and does not delay it.

PR #2535, opened 2018-07-17
Add `hideNotBuilt` and remove `showNotBuilt` named system code

deepikasl:
https://github.com/Shippable/www/issues/14443

Before:
screenshot from 2018-07-17 15 57 01

After:
screenshot from 2018-07-17 15 58 13

### DIFF
--- a/common/scripts/configs/system_codes.sql
+++ b/common/scripts/configs/system_codes.sql
@@ -775,9 +775,9 @@ do $$
       values (504, 'account', 'viewType', '54188262bc4d591ba438d62a', '54188262bc4d591ba438d62a', '2016-06-01', '2016-06-01');
     end if;
 
-    if not exists (select 1 from "systemCodes" where code = 505) then
+    if not exists (select 1 from "systemCodes" where code = 506) then
       insert into "systemCodes" ("code", "name", "group", "createdBy", "updatedBy", "createdAt", "updatedAt")
-      values (505, 'showNotBuilt', 'viewObjectType', '54188262bc4d591ba438d62a', '54188262bc4d591ba438d62a', '2018-06-25', '2018-06-25');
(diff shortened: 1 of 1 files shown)


common/scripts/configs/system_codes.sql
```diff
@@ -775,9 +775,14 @@ do $$
...
 
-    if not exists (select 1 from "systemCodes" where code = 505) then
+    if not exists (select 1 from "systemCodes" where code = 506) then
       insert into "systemCodes" ("code", "name", "group", "createdBy", "updatedBy", "createdAt", "updatedAt")
-      values (505, 'showNotBuilt', 'viewObjectType', '54188262bc4d591ba438d62a', '54188262bc4d591ba438d62a', '2018-06-25', '2018-06-25');
+      values (506, 'hideNotBuilt', 'viewObjectType', '54188262bc4d591ba438d62a', '54188262bc4d591ba438d62a', '2018-07-17', '2018-07-17');
+    end if;
+
+    -- Delete showNotBuilt systemCode
+    if exists (select 1 from information_schema.columns where table_name = 'systemCodes') then
```
shrivara: Is it safe to delete this? Won't there be any references to this code?

deepikasl: we are not using this anywhere.

deepikasl: I have added it before to add a option for show/hide not built branches. But we have decided to rename it to `hideNotBuilt` instead of `showNotBuilt`

shrivara: Is `showNotBuilt` already added in production?

shrivara: If it not we should just clean it up in RC and use 505.

deepikasl: yes its already added in production

shrivara: Okay, can we follow the below template for deleting the system codes? Looks good otherwise.

deepikasl: okay

PR #2587, opened 2018-08-07
explicitly export REL_VER

Bharath92:


shippable.yml
```diff
@@ -143,7 +143,7 @@ jobs:
                 - RES_REPO: "admiral_sh_repo"
                 - GIT_PATH: "git@github.com:Shippable/admiral.git"
           script:
-            - REL_VER=$(shipctl get_resource_version_key "$RES_VER" "versionName")
+            - export REL_VER=$(shipctl get_resource_version_key "$RES_VER" "versionName")
```
Bharath92: we need to explicitly export REL_VER for it to be substituted

PR #2594, opened 2018-08-10
adding admiral route to provide status

ric03uec:

### DIFF
--- a/api/admiral/getStatus.js
+++ b/api/admiral/getStatus.js
@@ -1,0 +1,93 @@
+'use strict';
+
+var self = getStatus;
+module.exports = self;
+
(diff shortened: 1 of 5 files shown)


api/admiral/getStatus.js
```diff
@@ -0,0 +1,93 @@
...
+          new ActErr(who, ActErr.OperationFailed,
+            'Script returned code: ' + exitCode)
+        );
+      return next();
+    }
+  );
+}
+
+function _getStatus(bag, next) {
+  if (!bag.resBody.isReachable) return next();
```
Bharath92: this is always false. probably, we should set it to true somewhere above?

PR #2596, opened 2018-08-13
printing admiral status

ric03uec:

### DIFF
--- a/common/scripts/status.sh
+++ b/common/scripts/status.sh
@@ -1,9 +1,42 @@
 #!/bin/bash
 set -e
 
+__parse_response() {
+  local api_response="$@"
+  if [ -z "$api_response" ]; then
+    __process_error "No response received from api, exiting"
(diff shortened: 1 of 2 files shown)


common/scripts/status.sh
```diff
@@ -1,9 +1,42 @@
...
+  printf "%2s COMPONENT %5s IP %9s PORT %10s UPTIME %8s REACHABLE \n"
+  printf "%10s %15s %10s %20s %10s \n" "$status_component" "$component_ip" \
+    "$component_port" "$component_uptime" "$component_rechable"
+
+  printf "\n"
+}
+
+__get_admiral_status() {
+  __process_marker "Checking admiral status"
+  _shippable_get_admiral_status
```
Bharath92: should we handle error status codes like this https://github.com/Shippable/admiral/blob/master/common/scripts/upgrade.sh#L38?

PR #2597, opened 2018-08-13
update db status route

ric03uec:

### DIFF
--- a/api/db/getStatus.js
+++ b/api/db/getStatus.js
@@ -5,12 +5,20 @@ module.exports = self;
 
 var async = require('async');
 var pg = require('pg');
+var _ = require('underscore');
+var configHandler = require('../../common/configHandler.js');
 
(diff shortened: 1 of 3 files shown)


api/db/getStatus.js
```diff
@@ -52,27 +89,71 @@ function _createClient(bag, next) {
...
+
+  exec.stdout.on('data',
+    function (data)  {
+      bag.resBody.uptime = data.toString();
+    }
+  );
+
+  exec.stderr.on('data',
+    function (data)  {
+      bag.resBody.error = data.toString();
```
Bharath92: should we error or out `bag.resBody.error` is not empty?

ric03uec: if this shows error, it'll be probably not be because db is down(we've already checked that in the previous step). so the error should ideally be propagated to the calller

PR #2611, opened 2018-08-22
upgrading docker on workers 

ric03uec:
- https://github.com/Shippable/pm/issues/11277
- docker on workers was not getting upgraded with the `./admiral.sh upgrade` command. this PR fixes the workflow
- this also refactors the workflow to install docker
    - the __check_dependencies() function will now **only** install docker if its not already present. Earlier this was installing docker unconditionally 
    - upgrade workflow will always run the install_docker scripts. this will make sure that proxy envs are correctly set. this was earlier taken care of by the __check_dependencies() function
- scenarios tested
    - fresh onebox install
    - multibox install
    - upgrade on onebox from old docker to new docker
    - upgrade on multi-box from old docker to new docker

### DIFF
--- a/common/scripts/upgrade.sh
+++ b/common/scripts/upgrade.sh
@@ -4,6 +4,13 @@ readonly MAX_ERROR_LOG_COUNT=100
 export IS_SERVER=false
 export system_cluster=null
 
+__update_docker() {
+  __process_msg "Updating docker"
+  ##if [ "$INSTALLED_DOCKER_VERSION" != "$DOCKER_VERSION" ]; then
+  source $SCRIPTS_DIR/upgrade_docker.sh
+  __process_msg "DONE"
+}
+
 __check_admiral() {
   __process_msg "Checking if admiral container is running"
   local wait_time=3
@@ -456,9 +463,7 @@ main() {
     __process_marker "Upgrading Shippable installation"
     __pull_images_master
     __pull_images_workers
-    if [ "$INSTALLED_DOCKER_VERSION" != "$DOCKER_VERSION" ]; then
-      source $SCRIPTS_DIR/upgrade_docker.sh
-    fi
+    __update_docker
     __check_admiral
     __check_release
     __update_release
--- a/common/scripts/upgrade_docker.sh
+++ b/common/scripts/upgrade_docker.sh
@@ -42,8 +42,10 @@ __remove_services() {
   fi
 }
 
-__upgrade_docker_version_on_swarm_workers() {
+__update_docker_on_swarm_workers() {
   __process_msg "Upgrading docker version on swarm workers"
+  ## Note: this always needs to execute since it'll update proxy settings on
+  ## swarm worker nodes
 
   local system_settings="PGPASSWORD=$DB_PASSWORD \
     psql \
@@ -112,7 +114,10 @@ __upgrade_docker_version_on_swarm_workers() {
   done
 }
 
-__upgrade_docker_version() {
+__update_docker() {
+  ## Note: this always needs to execute since it'll update proxy settings on
+  ## swarm master
+
   __process_msg "Upgrading docker to $DOCKER_VERSION"
   __create_install_docker_script
   ./$INSTALL_DOCKER_SCRIPT
@@ -153,8 +158,8 @@ main() {
   __process_marker "Upgrading docker"
 
   __remove_services
-  __upgrade_docker_version
-  __upgrade_docker_version_on_swarm_workers
+  __update_docker
+  __update_docker_on_swarm_workers
   __upgrade_awscli_version
   __set_installed_docker_version
   __restart_db_container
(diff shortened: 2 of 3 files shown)


common/scripts/upgrade.sh
```diff
@@ -4,6 +4,13 @@ readonly MAX_ERROR_LOG_COUNT=100
 export IS_SERVER=false
 export system_cluster=null
 
+__update_docker() {
```
Bharath92: this will affect other installations of docker as well. In U14, C7, RHEL the docker version will already be upgraded in the `_helpers.sh` file

common/scripts/upgrade_docker.sh
```diff
@@ -25,161 +42,124 @@ __remove_services() {
...
 }
 
 __restart() {
-  if $DOCKER_UPGRADE_REQUIRED; then
-    IS_RESTART=true
-    export skip_starting_services=true
-    source $SCRIPTS_DIR/restart.sh
-    IS_RESTART=false
-  fi
+  IS_RESTART=true
```
Bharath92: should this be done in every `upgrade` flow?

common/scripts/x86_64/Ubuntu_16.04/_helpers.sh
```diff
@@ -78,37 +78,48 @@ __check_dependencies() {
...
-  rm -rf $node_scripts_location
-  # Install Docker
-  chmod +x installDockerScript.sh
-  ./installDockerScript.sh
-
-  # sets env INSTALLED_DOCKER_VERSION
-  __set_installed_docker_version
-
-  rm installDockerScript.sh
+  {
```
Bharath92: this can just be
`check_docker=$(sudo service --status-all 2>&1 | grep docker) || true`

common/scripts/upgrade_docker.sh
```diff
@@ -25,161 +42,124 @@ __remove_services() {
...
-    cat $node_scripts_location/lib/headers.sh >> $INSTALL_DOCKER_SCRIPT
-    cat $node_scripts_location/initScripts/$ARCHITECTURE/$OPERATING_SYSTEM/Docker_"$DOCKER_VERSION".sh >> $INSTALL_DOCKER_SCRIPT
-
-    rm -rf $node_scripts_location
-    # Install Docker
-    chmod +x $INSTALL_DOCKER_SCRIPT
-  fi
-}
+__update_docker_on_swarm_workers() {
+  __process_msg "Upgrading docker version on swarm workers"
```
Bharath92: this will be new functionality. i dont think swarm workers' proxy settings were updated on every upgrade before. is this ok?

common/scripts/upgrade_docker.sh
```diff
@@ -25,161 +42,124 @@ __remove_services() {
...
 
 __restart() {
-  if $DOCKER_UPGRADE_REQUIRED; then
-    IS_RESTART=true
-    export skip_starting_services=true
-    source $SCRIPTS_DIR/restart.sh
-    IS_RESTART=false
-  fi
+  IS_RESTART=true
+  export skip_starting_services=true
```
Bharath92: this is not used anywhere currently

common/scripts/upgrade_docker.sh
ric03uec: we were supposed to do that. it doesnt make sense to just update master and leave the workers with old settings. i'd say this change is fixing a bug rather than introducing a new funtionality

common/scripts/x86_64/Ubuntu_16.04/_helpers.sh
ric03uec: then it'll be always true wont it?

common/scripts/upgrade_docker.sh
ric03uec: resetting docker proxy envs will restart docker anyway, which means its better to just restart services cleanly rather than letting docker handle them with defaults

common/scripts/upgrade.sh
ric03uec: we'll change that in next set of PR's

common/scripts/upgrade_docker.sh
ric03uec: this is being used in _restart.sh file

ric03uec:
the problem here is that we cant really skip running the docker installation script
- we need to update proxy envs for docker daemon every single time user runs upgrade
- proxy envs for docker daemon are only getting set in docker install scripts
- so we need to run docker install scripts with each upgrade, whether or not docker is installed on the box(swarm master/workers) or not

common/scripts/upgrade_docker.sh
Bharath92: ok

common/scripts/x86_64/Ubuntu_16.04/_helpers.sh
Bharath92: no it will be true if the command `sudo service --status-all 2>&1 | grep docker` fails

Bharath92: this is incorrect. Merging this based on the conversation on slack with @ric03uec  that this is done according to the conventions followed previously in admiral.

Issue #2629, opened 2018-09-03
11304.2 DEV update post and put systemIntegration routes to create and add ssh keys to gerritBasicAuth systemIntegration


ankul-shippable:
Shippable/pm#11304

ankul-shippable:
this is done, closing.

PR #2632, opened 2018-09-04
 update gerrit auth provider to add credentials for gerrit event-streams

ankul-shippable:
https://github.com/Shippable/admiral/issues/2629

### DIFF
--- a/api/systemIntegrations/post.js
+++ b/api/systemIntegrations/post.js
@@ -6,6 +6,8 @@ module.exports = self;
 var async = require('async');
 var mongoose = require('mongoose');
 var _ = require('underscore');
+var tmp = require('tmp');
+var sshKeyGen = require('ssh-keygen');
 
 var APIAdapter = require('../../common/APIAdapter.js');
 var envHandler = require('../../common/envHandler.js');
@@ -33,6 +35,7 @@ function post(req, res) {
       _getMasterIntegrationFields.bind(null, bag),
       _validateMasterIntegrationFields.bind(null, bag),
       _postProvider.bind(null, bag),
+      _genSshKeys.bind(null, bag),
       _createSystemIntegration.bind(null, bag),
       _getSystemIntegration.bind(null, bag),
       _postSystemIntegrationFieldsToVault.bind(null, bag)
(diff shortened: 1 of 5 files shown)


api/systemIntegrations/post.js
```diff
@@ -274,6 +271,14 @@ function _postProvider(bag, next) {
     name = name.replace('Keys', '');
   else if (name.endsWith('BasicAuth'))
     name = name.replace('BasicAuth', '');
+  bag.providerName = name;
+
+  if (!bag.reqBody.data.url) {
```
rageshkrishna: Shouldn't these checks happen in checkInputParams?

api/systemIntegrations/put.js
```diff
@@ -246,6 +243,14 @@ function _postProvider(bag, next) {
     name = name.replace('Keys', '');
   else if (name.endsWith('BasicAuth'))
     name = name.replace('BasicAuth', '');
+  bag.providerName = name;
+
+  if (!bag.reqBody.data.url) {
```
rageshkrishna: Shouldn't these checks happen in checkInputParams?

api/systemIntegrations/put.js
```diff
@@ -265,6 +270,35 @@ function _postProvider(bag, next) {
   );
 }
 
+function _genSshKeys(bag, next) {
```
rageshkrishna: Why is this necessary during the PUT? If the keys were not generated, the integration would not have been created to begin with, right?

api/systemIntegrations/post.js
ankul-shippable: this is added here because initially we create the sysInt without any url. so if the url is not present then we dont create the provider. If it is present, then we create the provider.

api/systemIntegrations/put.js
ankul-shippable: this is added here because initially we create the sysInt without any url. so if the url is not present then we dont create the provider. If it is present, then we create the provider.

api/systemIntegrations/put.js
ankul-shippable: ok, removing it.

api/systemIntegrations/post.js
rageshkrishna: Ok, makes sense. I misread the diff.

Issue #2630, opened 2018-09-03
11304.4 DEV update admiral to start gerritEventHandler service if gerrit auth providers are added


ankul-shippable:
Shippable/pm#11304

ankul-shippable:
this is done, closing.

PR #2635, opened 2018-09-06
start and stop gerrit event handler service based on gerrit auth sysInts

ankul-shippable:
https://github.com/Shippable/admiral/issues/2630


static/scripts/dashboard/dashboardNewCtrl.js
```diff
@@ -4890,6 +4891,34 @@
       }
     }
 
+    function startOrStopGerritEventHandler(next) {
+      var intercomProcService = _.findWhere($scope.vm.allServices,
```
rageshkrishna: Wrong var name, but why do you even need this?

PR #2664, opened 2018-09-19
adds release notes regarding RHEL7 installation bug

Bharath92:
https://github.com/Shippable/heap/issues/2553


ric03uec:
lets link onebox text with http://docs.shippable.com/platform/server/install-onebox/
we should try to link the release notes back to docs and/or blogs

Issue #2743, opened 2018-11-23
add the new systemCodes for status view


ChaitanyaVootla:
https://github.com/Shippable/pm/issues/11569

PR #2744, opened 2018-11-23
adds system codes for status view

ChaitanyaVootla:
https://github.com/Shippable/admiral/issues/2743


rageshkrishna:
Please move all of this to the bottom of the file. We should not introduce new migrations in between.

Issue #2874, opened 2019-03-01
 Correct airBrakeKey masterIntegration level and display name


scriptnull:
https://github.com/Shippable/pm/issues/11780

scriptnull:
Verified on RC

PR #2875, opened 2019-03-01
updates airBrakeKey level and displayName

scriptnull:
https://github.com/Shippable/admiral/issues/2874

Following queries are already run on RC Database.

```
delete from "masterIntegrationFields" where "masterIntegrationId" = '577de63321333398d11a1124';
delete from "masterIntegrations" where id = '577de63321333398d11a1124';
```


scriptnull:
I guess @rageshkrishna missed out this and seems to be unavailable now. I am going ahead and merging since we need to ensure that the already present masterIntegration and masterIntegrationFields are deleted from the database while this change is deployed.

If we delete now and wait for merge, any intermittent deployment to RC will bring back those entries.

